PR #1, opened 2019-07-11
Operator now supports device files outside the /dev

aravindavk:
- If device file is specified is outside `/dev` then Operator will
  mount the parent directory of file to server container.
- Incremented the Fedora version to 30 so that it will pull the latest
  stable gluster version(6.3)
- `include_uninitialized` argument is removed in latest kubernetes python
  client. So removed from Operator code.

### DIFF
--- a/templates/server.yaml.j2
+++ b/templates/server.yaml.j2
@@ -68,9 +68,13 @@ spec:
               readOnly: true
             - name: glusterfsd-volfilesdir
               mountPath: "/var/lib/gluster"
-            - name: glusterfsd-datadir
-              mountPath: "/bricks"
+            - name: glusterfsd-mountdir
+              mountPath: "/bricks/{{ volname }}"
               mountPropagation: Bidirectional
+{%- if brick_device_dir != "" %}
+            - name: brick-device-dir
+              mountPath: "/brickdev"
+{%- endif %}
         - name: quotad
           image: docker.io/{{ docker_user }}/kadalu-server:{{ kadalu_version }}
           env:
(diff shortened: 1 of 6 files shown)


templates/server.yaml.j2
```diff
@@ -97,6 +101,8 @@ spec:
             - name: glusterfsd-datadir
               mountPath: "/bricks"
               mountPropagation: Bidirectional
+            - name: glusterfsd-volfilesdir
+              mountPath: "/var/lib/gluster"
```
amarts: What is this? Why is this required?

templates/server.yaml.j2
```diff
@@ -111,10 +117,16 @@ spec:
...
+{%- if brick_device == "" %}
           hostPath:
             path: "{{ host_brick_path }}"
             type: Directory
-{% else %}
+{%- else %}
           emptyDir: {}
-{% endif %}
+{%- endif %}
+{%- if brick_device_dir != "" %}
```
amarts, line 125: Looks like we need to get the similar logic for 'path' too. Otherwise, path directory is not exported inside glusterfsd container IMO.

templates/server.yaml.j2
aravindavk: I saw a log from quotad container, let me recheck. May be a setup issue. If it works without this change I will remove.

templates/server.yaml.j2
aravindavk: It is already handled for path directory https://github.com/aravindavk/kadalu/blob/master/templates/server.yaml.j2#L114 

Looks like not working as expected, I will debug further.

aravindavk: Silly mistake... used `or` instead of `and` here https://github.com/aravindavk/kadalu/blob/master/server/glusterfsd.py#L117

amarts: Cool :+1

aravindavk:
Updated the code and tested with all three cases. Please review

- Provide `device: /dev/vdc` or `device: /dev/loop0`
- Provide `device: /bricks/brick1.img`
- Provide `path: /bricks/brick2`

amarts (approved):
LGTM

PR #13, opened 2019-07-25
csi: make sure glusterfs logs are sent to stdout

### LINKED ISSUE #9: Centralized logging
k8s takes care of having all '/dev/stdout' logs in a centralized format. Currently glusterfsd logs are all going to stdout, but the CSI logs are not going to /dev/stdout. Need to get them in central place, so debugging would be easy.

amarts:
Fixes: #9

### DIFF
--- a/csi/volumeutils.py
+++ b/csi/volumeutils.py
@@ -427,6 +427,7 @@ def mount_glusterfs(volume, target_path):
     cmd = [
         GLUSTERFS_CMD,
         "--process-name", "fuse",
+        "-l", "/dev/stdout",
         "--volfile-id=%s" % volume,
         target_path,
         "-f", "%s/%s.client.vol" % (VOLFILES_DIR, volume)


csi/volumeutils.py
```diff
@@ -427,6 +427,7 @@ def mount_glusterfs(volume, target_path):
     cmd = [
         GLUSTERFS_CMD,
         "--process-name", "fuse",
+        "-l", "/dev/stdout",
```
aravindavk, line 430: stdout of this child process is not consumed.

amarts: 'stdout' (ie, fd=1) is not consumed, but here, glusterfs opens `-l /dev/stdout` as any other file. So, it would be writing to stdout device even faster

amarts:
This is good to try IMO.

aravindavk (approved):

PR #14, opened 2019-07-25
csi: make gluster mount with --auto-invalidation off

### LINKED ISSUE #11: consider using fuse's write-cache on CSI mount.
Fuse's write-cache would give much better performance than that of glusterfs's internal caching. It would be good to provide that here.

amarts:
This makes use of kernel's fuse cache.

Refer: `git show a229ee1c8cdf8e0ac1abaeb60cabe6ab08f60546` for more
info on this. considering each directory is handled differently, this
may make sense here.

Tag: Experimental

Fixes: #11

### DIFF
--- a/csi/volumeutils.py
+++ b/csi/volumeutils.py
@@ -427,6 +427,7 @@ def mount_glusterfs(volume, target_path):
     cmd = [
         GLUSTERFS_CMD,
         "--process-name", "fuse",
+        "-l", "/dev/stdout",
         "--volfile-id=%s" % volume,
         target_path,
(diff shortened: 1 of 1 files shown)


amarts:
No need to merge this immediately. Lets merge this after tagging v1.

amarts:
Looks like there is some issue with option. Even though it is a good answer to performance of gluster, we may need to have it as option. For now, this PR can wait.

amarts:
Noticed that this is dependent on host machine capabilities. The impact is not yet tested. Hence marking it not required.

Issue #16, opened 2019-07-26
Analytics of usage


amarts:
It would be great to add analytics on usage (specially just the downloads), and details on which version it uses etc. With this information, it is easier to support users, and also provide features in particular version.

PR #18, opened 2019-07-26
enable tracker for operator and server

amarts:
* Tracks 1 page view before starting the server
* Tracks 1 page view after starting operator

Nothing other than page view is tracked. This helps us to enhance
supportability in future by adding k8s version etc in tracker.

Also to be worked on, to make it optional.

Updates: #16


aravindavk (approved):

amarts:
Can we build a container with this? Lets keep it for few days, and see if there is any tracking at all. If nothing, we can remove it. If it works good, we can publish the results in README every month.

PR #21, opened 2019-07-29
minikube.sh: use copy_image_to_cluster method before starting tests

amarts:
Make sure that KADALU_VERSION is not 'latest' when testing.

### DIFF
--- a/tests/minikube.sh
+++ b/tests/minikube.sh
@@ -15,19 +15,6 @@ function wait_for_ssh() {
     exit 1
 }
 
-function copy_image_to_cluster() {
-    local build_image=$1
-    local final_image=$2
-    if [ -z "$(docker images -q "${build_image}")" ]; then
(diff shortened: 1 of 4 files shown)


tests/minikube.sh
```diff
@@ -105,7 +103,12 @@ kadalu_operator)
     # TODO: need to use the locally built images, so we can test changes to Dockerfiles too
     echo "Starting the kadalu Operator"
 
+    copy_image_to_cluster "${KADALU_IMAGE_REPO}"/operator:canary "${KADALU_IMAGE_REPO}"/operator:canary
```
aravindavk: kadalu-operator

amarts: Ack. Fixed in later versions.

amarts:
Considering this is critical for all the tests in future, merging this.

PR #39, opened 2019-10-02
tracker: add a client id based on timestamp

amarts:
* Also fix README commit link which showed the tracker we added.


amarts:
Updates #16

PR #44, opened 2019-10-07
volfile-template: remove distribute and io-cache from the template

amarts:


aravindavk (approved):
Looks good. When Dist replicate is supported we need to add new volfiles

Issue #49, opened 2019-10-25
Documentation for gluster part missing


papanito:
I read trough your documents but for me it is still not clear what is required or has to be done in terms of gluster. 
Does glusterd need to be running on the host? What are the pre-requirements?  

In your blog entry you say

> expose the storage in one of the nodes of k8s, and provide storage config file to k8s.

Maybe you can elaborate this a bit more.

Thanks

aravindavk:
Running Glusterd is not required on the host. Once Kadalu Operator is deployed, a Storage device from the Host needs to be specified.

Below Asciinema recordings may help to understand the setup and exposing the storage device.

- Setup: https://asciinema.org/a/259949
- Exposing the storage device: https://asciinema.org/a/259951

Hope this helps.

We are working on improving the documentation. Thanks for reporting the issue.

amarts:
@papanito I hope Aravinda's answers gave some insights to you.

>  what is required or has to be done in terms of gluster.

Provide a storage (or 3 storage if it is replica 3) end points for kadalu to export. (ie, brick detail in gluster terminology). Gluster will be setup automatically, no need to do anything other than telling which device to export.

> Does glusterd need to be running on the host? 

If you go through our fundamental idea of the project [(in pdf here)](https://github.com/kadalu/kadalu/blob/master/doc/rethinking-gluster-management-using-k8s.pdf) There is no need to have a 'glusterd'. That is the beauty of this.

> What are the pre-requirements?

There are few:

* Kubernetes 1.14.0 + version
* The host should support xfs (mkfs.xfs)
* The `mount  -t xfs` with `-oprjquota` should work (noticed recently that on minikube VM this doesn't work).

And thats it :-)

amarts:
Also Issue #42 has some questions answered.

papanito:
Thanks guys thus helps me a lot. Will give it a try

amarts:
Closed with #50

PR #50, opened 2019-10-28
#49 Improve documentation

papanito:
Based on your feedback in my issue, I figured it make sense to write that down in the documentation. Maybe not the final structure, but at least information is more visible to the user


amarts (approved):

aravindavk (approved):
Looks good to me. Thanks for the PR

amarts:
Just want to comment one thing here for future completeness.

We deliberately kept Gluster documentation from the front page thinking, for k8s users/admins, they needn't bother about Gluster while using Kadalu :-)

But, @papanito's thoughts are also valid, those who know gluster, and searching for heketi alternative, may need to read about gluster part to understand the solution. Till we get to better documentation, these information is good for everyone :+1:

papanito:
Thanks guys!

PR #62, opened 2019-11-13
Deploy Attacher as container in  provisioner statefulset

fs-system:
Deploy attacher as  a container in provisioner statefulset instead  of deploying it as a separate statefulset

This PR also adds functionality to create CSIDriver object which will  help people to list the CSI drivers deployed in theier environment

PR #63, opened 2019-11-19
tests: update the test scripts, and add test for k8s 1.16.0

amarts:
Updates: #62

Credits: @fs-system

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -25,7 +25,7 @@ env:
     - GO_METALINTER_THREADS=1
     - GO_COVER_DIR=_output
     - VM_DRIVER=none
-    - MINIKUBE_VERSION=v1.2.0
+    - MINIKUBE_VERSION=v1.5.2
     - CHANGE_MINIKUBE_NONE_USER=true
     - KUBECONFIG=$HOME/.kube/config
 
@@ -36,3 +36,8 @@ jobs:
         - make build-containers || travis_terminate 1;
         - make gen-manifest || travis_terminate 1;
         - tests/travis-test.sh v1.15.0 || travis_terminate 1;
+    - name: kadalu with kube 1.16.0
+      script:
+        - make build-containers || travis_terminate 1;
+        - make gen-manifest || travis_terminate 1;
+        - tests/travis-test.sh v1.16.0 || travis_terminate 1;
--- a/tests/minikube.sh
+++ b/tests/minikube.sh
@@ -15,6 +15,21 @@ function wait_for_ssh() {
     exit 1
 }
 
+function copy_image_to_cluster() {
+    local build_image=$1
+    local final_image=$2
+    if [ -z "$(docker images -q "${build_image}")" ]; then
+        docker pull "${build_image}"
+    fi
+    if [[ "${VM_DRIVER}" == "none" ]]; then
+        docker tag "${build_image}" "${final_image}"
+        return
+    fi
+    docker save "${build_image}" | \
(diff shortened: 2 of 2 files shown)


manifests/kadalu-operator.yaml
```diff
@@ -192,7 +192,7 @@ spec:
             capabilities: {}
             privileged: true
           image: docker.io/kadalu/kadalu-operator:latest
-          imagePullPolicy: Always
+          imagePullPolicy: IfNotPresent
```
fs-system: This is not required, i think we need to remove this

.travis.yml
```diff
@@ -36,3 +36,8 @@ jobs:
         - make build-containers || travis_terminate 1;
         - make gen-manifest || travis_terminate 1;
         - tests/travis-test.sh v1.15.0 || travis_terminate 1;
+    - name: kadalu with kube 1.16.0
```
fs-system, line 39: can you add Kube 1.13 also?

amarts: I will add 1.13 as a separate patch. Lets take one at a time :-)

And also, I believe for 1.13.7 we need to add some extra checks, and different files (object/crd) in manifest?

fs-system: nope, nothing is required, current code should work with kube 1.13 as well.

manifests/kadalu-operator.yaml
amarts: Done

.travis.yml
amarts: Ok, let me try that as a patch on top of this. Let me get to fix this.

amarts:
@aravindavk @fs-system Check if these changes are fine, and merge, so I can start adding k8s 1.13.7 version also to the mix.

tests/minikube.sh
```diff
@@ -127,72 +157,84 @@ kadalu_operator)
...
+
+    echo "---- Logs from pod2 ----"
+    kubectl describe pod pod2
+    kubectl logs pod2
+    echo "---- Logs from pod1 ----"
+    kubectl describe pod pod1
+    kubectl logs pod1
     #kubectl -n kadalu logs -lapp=kadalu --all-containers=true
 
+    date
```
fs-system, line 232: remove this?

tests/minikube.sh
```diff
@@ -127,72 +157,84 @@ kadalu_operator)
...
-        if [[ $((cnt%10)) -eq 0 ]]; then
+        if [[ $((cnt % 10)) -eq 0 ]]; then
             echo "$cnt: Waiting for pods to come up..."
         fi
     done
-    kubectl logs pod2
 
     # Log everything so we are sure if things are as expected
     for p in $(kubectl -n kadalu get pods -o name); do
-        echo "====================== Start $p ======================";
```
fs-system, line 187: i think we can enable logging only if test fails

amarts: I wanted log there, so in worst case, we can validate if things are OK or not OK just by looking at even successful test.

Anyways, if the tests are passing, we wouldn't bother to check the tests and would look at only code changes. Only when it fails, we goto CI/CD system. I am inclined to keep it.

tests/minikube.sh
amarts: This would help us to relate the logs w.r.to the start (or end) time. Hence added as extra debugging. We can remove them eventually. For now extra debugging info (in what so ever form) would help us.

Hope thats fine?

fs-system:
@amarts we can merge this one

PR #64, opened 2019-11-19
K8s test for 1.13.7

amarts:


amarts:
This too, like #63, is an extended breakup from #62 

Credits @fs-system

Issue #67, opened 2019-11-21
Add CI/CD for replica3 volume


amarts:
It is not possible to make sure we have 100% usecases covered manually for each PR. If a case is not covered in CI/CD, the usecase can't be supported.

We need to make sure we also test Replica3 in CI/CD. Basic things shouldn't come as bug to users like #65

amarts:
```
[2019-11-21 17:38:28,904] DEBUG [controllerserver - 42:CreateVolume] - Filters applied to choose storage	 hostvol_type=Replica1
[2019-11-21 17:38:28,905] DEBUG [controllerserver - 47:CreateVolume] - Got list of hosting Volumes	 volumes=storage-pool-2,storage-pool-1
```
Here, storage-pool-2 should have been Replica3 type.

PR #68, opened 2019-11-21
tests: add more cases

amarts:
TODO: make sure RWO/RWX both are tested on both 'replica1' and 'replica3' types

Updates: #67


amarts:
Prefer 'rebase and merge' as the latest commit is proper and doesn't have a TODO.

operator/main.py
```diff
@@ -108,11 +108,11 @@ def get_brick_hostname(volname, node, suffix=True):
     # Service is created with name as Volume name. For example,
     # brick_hostname will be "server-spool1-minikube.spool1" and
     # server pod name will be "server-spool1-minikube"
-    hostname = "server-%s-%s" % (volname, node.replace(".", "-"))
+    hostname = "server-%s-%d-%s" % (volname, idx, node.replace(".", "-"))
     if suffix:
         # TODO: ordinal is used as zero. If more than one brick used from the
         # same node and for the same volume then consider changing this
-        return hostname + "-0." + volname
+        return "%s-%d.%s" % (hostname, idx, volname)
```
aravindavk: This will go wrong in multi node setup. suffix is assigned by K8s. If brick pod is started in other node it will still have `-0` unless more than one brick with same pod name.

I think we can change this to `"%s-0.%s" % (hostname, volname)`  Note: idx is already part of hostname in line 111

Above TODO can be removed after this change.

amarts: Done

aravindavk (approved):

Issue #65, opened 2019-11-21
Failed to provision volume with StorageClass "kadalu.replica3"


hetii:
Hi,
I try to setup kadalu on aks cluster but csi-provisioner raise error like:

    Warning  ProvisioningFailed    3s                 kadalu_csi-provisioner-0_27e261cc-0c6d-11ea-a863-7ec92e551436  failed to provision volume with StorageClass "kadalu.replica3": rpc error: code = Unknown desc = Exception calling application: [Errno 17] File exists: '/mnt/kadalu-storage-replica-pool-3'

Details:

    I1121 14:50:40.022459       1 controller.go:926] provision "default/kadalu-pvc" class "kadalu.replica3": started
    I1121 14:50:40.028583       1 event.go:221] Event(v1.ObjectReference{Kind:"PersistentVolumeClaim", Namespace:"default", Name:"kadalu-pvc", UID:"40c1247f-75dd-43da-b304-ed200e5874dd", APIVersion:"v1", ResourceVersion:"238587", FieldPath:""}): type: 'Normal' reason: 'Provisioning' External provisioner is provisioning volume for claim "default/kadalu-pvc"
    W1121 14:50:40.033254       1 controller.go:685] Retrying syncing claim "default/kadalu-pvc" because failures 5 < threshold 15
    E1121 14:50:40.033333       1 controller.go:700] error syncing claim "default/kadalu-pvc": failed to provision volume with StorageClass "kadalu.replica3": rpc error: code = Unknown desc = Exception calling application: [Errno 17] File exists: '/mnt/kadalu-storage-replica-pool-3'
    I1121 14:50:40.033410       1 event.go:221] Event(v1.ObjectReference{Kind:"PersistentVolumeClaim", Namespace:"default", Name:"kadalu-pvc", UID:"40c1247f-75dd-43da-b304-ed200e5874dd", APIVersion:"v1", ResourceVersion:"238587", FieldPath:""}): type: 'Warning' reason: 'ProvisioningFailed' failed to provision volume with StorageClass "kadalu.replica3": rpc error: code = Unknown desc = Exception calling application: [Errno 17] File exists: '/mnt/kadalu-storage-replica-pool-3'

    [2019-11-21 14:58:40,050] DEBUG [controllerserver - 42:CreateVolume] - Filters applied to choose storage   hostvol_type=Replica3
[...]

PR #69, opened 2019-11-22
Handle EEXIST error when hosting vol mount is reused

aravindavk:
`os.makedirs` handles EEXIST with additional flag `exist=True` but
fails if directory attributes are different. Added wrapper function
to handle EEXIST error gracefully.

Updates: #65


amarts (approved):
LGTM

PR #70, opened 2019-11-22
Start Self Heal daemon if the storage type is Replica3

### LINKED ISSUE #40: Start 'shd' container if 'Replica3' is choosen type.
Currently in `server/server.py` there is no self-heal daemon started. Would be great to have it running.

aravindavk:
Start self heal daemon in each server pod if the storage type
is specified as Replica3.

Fixes: #40

### DIFF
--- a/operator/main.py
+++ b/operator/main.py
@@ -139,7 +139,7 @@ def update_config_map(core_v1_client, obj):
         data["bricks"].append({
             "brick_path": "/bricks/%s/data/brick" % volname,
             "node": get_brick_hostname(volname, storage["node"], idx),
-            "node_id": str(uuid.uuid1()),
+            "node_id": storage["node_id"],
             "host_brick_path": storage.get("path", ""),
             "brick_device": storage.get("device", ""),
             "brick_device_dir": get_brick_device_dir(storage),
@@ -168,7 +168,8 @@ def deploy_server_pods(obj):
         "kadalu_version": VERSION,
         "docker_user": docker_user,
         "volname": volname,
-        "volume_id": obj["spec"]["volume_id"]
+        "volume_id": obj["spec"]["volume_id"],
+        "shd_required": obj["spec"]["type"] == VOLUME_TYPE_REPLICA_3
     }
 
     # One StatefulSet per Brick
@@ -186,6 +187,7 @@ def deploy_server_pods(obj):
         template_args["brick_index"] = idx
         template_args["brick_device"] = storage.get("device", "")
         template_args["brick_device_dir"] = get_brick_device_dir(storage)
+        template_args["brick_node_id"] = storage["node_id"]
 
         filename = os.path.join(MANIFESTS_DIR, "server.yaml")
         template(filename, **template_args)
@@ -219,6 +221,10 @@ def handle_added(core_v1_client, obj):
 
     # Generate new Volume ID
     obj["spec"]["volume_id"] = str(uuid.uuid1())
+
+    # Generate Node ID for each storage device.
+    for idx, storage in enumerate(obj["spec"]["storage"]):
+        obj["spec"]["storage"][idx]["node_id"] = str(uuid.uuid1())
 
     update_config_map(core_v1_client, obj)
     deploy_server_pods(obj)
--- a/server/glusterfsd.py
+++ b/server/glusterfsd.py
@@ -152,13 +152,11 @@ def start():
             "-N",
             "--volfile-id", volfile_id,
             "-p", "/var/run/gluster/glusterfsd-%s.pid" % brick_path_name,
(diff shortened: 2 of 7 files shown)


operator/main.py
```diff
@@ -141,7 +141,7 @@ def update_config_map(core_v1_client, obj):
         data["bricks"].append({
             "brick_path": "/bricks/%s/data/brick" % volname,
             "node": get_brick_hostname(volname, storage["node"]),
-            "node_id": str(uuid.uuid1()),
+            "node_id": storage["node_id"],
```
amarts, line 142: this is changed now? (or should be a conflict somewhere here..)

server/glusterfsd.py
```diff
@@ -152,13 +152,11 @@ def start():
...
             "-p", "/var/run/gluster/glusterfsd-%s.pid" % brick_path_name,
-            # TODO: Change socket file name
-            "-S", "/var/run/gluster/b99981c29a4c396c.socket",
+            "-S", "/var/run/gluster/brick.socket",
             "--brick-name", brick_path,
             "-l", "-",  # Log to stderr
-            # TODO: Change Node ID
             "--xlator-option",
-            "*-posix.glusterd-uuid=6958dddc-1842-4ee0-92df-b6a060dfba5e",
+            "*-posix.glusterd-uuid=%s" % os.environ["NODEID"],
```
amarts, line 159: NOTE TO SELF: need to understand this better (for myself). option has 'glusterd' in it, and we should understand why and what it does.

aravindavk:
Testing and CI changes yet to be done

server/shd.py
```diff
@@ -0,0 +1,64 @@
...
+    """
+    volname = os.environ["VOLUME"]
+
+    volfile_path = os.path.join(VOLFILES_DIR, "glustershd.vol")
+    generate_shd_volfile(volfile_path, volname)
+
+    # Send Analytics Tracker
+    # The information from this analytics is available for
+    # developers to understand and build project in a better way
+    send_analytics_tracker("server")
```
amarts: considering this is another container in server pod, we can skip this IMO

server/shd.py
```diff
@@ -0,0 +1,64 @@
...
+
+
+def generate_shd_volfile(volfile_path, volname):
+    """Generate Client Volfile for Glusterfs Volume"""
+    info_file_path = os.path.join(VOLINFO_DIR, "%s.info" % volname)
+    data = {}
+    with open(info_file_path) as info_file:
+        data = json.load(info_file)
+
+    template_file_path = os.path.join(TEMPLATES_DIR, "Replica3.shd.vol.j2")
```
amarts, line 27: with this feature, should we disable client self-heal altogether? or is it ok to keep it for couple of releases?

server/shd.py
aravindavk: ack

server/glusterfsd.py
aravindavk: May be to identify local node ID

server/shd.py
aravindavk: I am not sure, we can keep it if no perf penalty.

aravindavk:
Need to think about the support for Self heal status and trigger self heal without having CLIs

server/shd.py
amarts: There is some perf penalty for sure, I am not aware at present, how much.

amarts:
> Need to think about the support for Self heal status and trigger self heal without having CLIs

How about a tool which deploys a container, and does `find` on mount point on PV (or on the whole volume itself). Like our sample-app-test container, which just does read/write, and lists mount details ?

amarts:
Some thoughts:

Or a future sidecar container, which would be listening on a port, which can have a button from prometheus/grafana dashboard to trigger all this?

Or similar to current quotad.py which keeps watching for a file, and depending on the content, does some actions? like self-heal, like rebalance (future), or like a md5sum data check, etc etc?

amarts:
```

 File "/usr/local/lib64/python3.7/site-packages/jinja2/parser.py", line 59, in fail
    raise exc(msg, lineno, self.name, self.filename)
jinja2.exceptions.TemplateSyntaxError: Unexpected end of template. Jinja was looking for the following tags: 'elif' or 'else' or 'endif'. The innermost block that needs to be closed is 'if'.

```

Also operator crashlooped 4 times, and there are 'stale mount' warning.

aravindavk:
Still failing on CI. Checking

aravindavk:
Tests are passing now. Had an issue with shd volfile and server YAML files.

Tests are not covering the actual heal functionality, we may have to add those tests. I will see how to simulate the pod failures and testing self-heal later.

amarts (approved):

Issue #46, opened 2019-10-23
RFE: make kadalu-server run on top of RWO PV itself.


amarts:
Most of the cloud providers provide a decent RWO PV support. Can kadalu-server run on top of the RWO PV, instead of having a need for admin manually setup a block device, or an export directory.

amarts:
Addresses one of the usecases mentioned in #42

amarts:
https://kubernetes.io/docs/concepts/storage/volumes/#awselasticblockstore 

If you notice, the PV here can take `fsType` parameter, with which, we can make our storage work on top of it. All it needs it have the proper config parameters to the PVC saying we need xfs.

amarts:
above suggestion answers https://github.com/kadalu/kadalu/issues/42#issuecomment-545769621

hetii:
I just add that in azure is possible to define fsType by using such storage class:

    ---
    kind: StorageClass
    apiVersion: storage.k8s.io/v1
    metadata:
      name: premium-azure-disk-sc
    provisioner: kubernetes.io/azure-disk
    reclaimPolicy: Retain
    # reclaimPolicy: Delete
    allowVolumeExpansion: true
    volumeBindingMode: Immediate
    parameters:
      storageaccounttype: Premium_LRS
      kind: Managed
      fsType: xfs
      #kind: Shared

amarts:
It is tested on top of hostpath PVC of minikube right now, and added CI tests for the same.  The features should be available with v0.4 version.

amarts:
@hetii our v0.4 release is now out, and this feature is available to try. Just make sure that the host PVC is available for `kadalu` namespace.

PR #73, opened 2019-12-02
operator: deploy storage pod on top of PVC

amarts:
Work-In-Progress

For this, one needs to provide PVC name as argument in storage config, something like below.

```
---
kind: KadaluStorage
metadata:
 # This will be used as name of PV Hosting Volume
  name: storage-pool-1
spec:
  type: Replica1
  storage:
    - pvc: pvc-from-other-csi

```

TODO: Considering this gets deployed on 'any' node where CSI is available, need to see
how to get node details back into client volfile.

Updates: #46

### DIFF
--- a/operator/main.py
+++ b/operator/main.py
@@ -44,6 +44,10 @@ def bricks_validation(bricks):
     """Validate Brick path and node options"""
     ret = True
     for idx, brick in enumerate(bricks):
+        if brick.get("pvc", None) is not None:
+            ret = True
+            continue
+
         if brick.get("path", None) is None and \
            brick.get("device", None) is None:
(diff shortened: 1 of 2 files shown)


amarts:
upon checking further, looks like this should be good enough, as 'serverpod_name' is used as `remote-host` in volfile and that should resolve to proper IP IMO.

@hetii is this possible to confirm this? I may not be able to spend time on testing this for next 15days. Can get back to this only after, hence the request.

amarts:
@aravindavk I am thinking of merging this, and having it in v0.4. But not calling out it as 'ready'. We can call it an 'alpha' feature. Once we get feedback, we can call it ready by v0.5

aravindavk (approved):

PR #76, opened 2019-12-13
Improvement to test script and other minor improvements.

amarts:
* Improved the tests to finish in shorter time as all the PVC
  requests are now sent together (saves 200 seconds)
* add volume-type in context to be passed to NodePlugin
* make each sample PVC size as different (so we can validate quickly by looking at logs)
* add locking to prevent multiple host volume mount request in parallel


aravindavk:
I see one possible race condition when two requests see file not exists and try to create and one crashes during mknod instead of retry. This should be okay since request will be retried, But we may need to fix it using threading.Lock (https://docs.python.org/3/library/threading.html#lock-objects)

PR #81, opened 2019-12-13
working external storage

amarts:
How this works:

Gluster Host, and volume name are passed to kadalu through
storage class parameters, with a keyword 'External'.

If this keyword is found, then we just try to find if the volume
exists, continue with using the volume as PV.

Passing gluster options to mount point is not yet complete, and
would be completed later depending on the need.

### DIFF
--- a/README.md
+++ b/README.md
@@ -140,6 +140,8 @@ pod1 created
 
 For more information checkout the [Try it out](./doc/README.md)
 
+If you want 'External' Gluster Storage to be used as PV, checkout [this doc](./doc/external-gluster-storage.md)
+
 ## NOTE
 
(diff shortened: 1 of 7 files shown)


amarts:
From the tests :-)

```
This is a sample application
# df -h
Filesystem                Size      Used Available Use% Mounted on
gluster1.kadalu.io:test
                        100.0M         0    100.0M   0% /mnt/pv
# mount
gluster1.kadalu.io:test on /mnt/pv type fuse.glusterfs (rw,relatime,user_id=0,group_id=0,default_permissions,allow_other,max_read=131072)
Write/Read test on PV mount
Fri Dec 13 11:17:44 UTC 2019
SUCCESS
```

aravindavk (approved):

PR #85, opened 2019-12-17
Improved tests

amarts:
* call one PVC after another for better debugging.
* Provide both `path` and `pvc` option along with `device` option to storage config.


amarts:
this provides test for #73

PR #82, opened 2019-12-14
Kadalu external storage

amarts:
The logic is simple. External volume details are passed as part of
storage-class itself. Instead of applying filters to find the hosting
volume, if the host-vol type passed is 'External-Kadalu' in storageclass,
we use the same volume for creating subvols.

Ideally, we could have better control if PVC request can have these
details, but the dynamic options are not allowed with PVC, but are
part of StorageClass.

But this is good enought to get the external storage in action for
kadalu users.

Check `examples/sample-external-kadalu-storage.yaml` for getting the
idea.

Updates #36

PR #88, opened 2019-12-18
kadalu with external gluster cluster

amarts:
Check both the commit message for better understanding of what is being done.


amarts:
Merging with @aravindavk's approval for #82

PR #89, opened 2019-12-18
kubectl-kadalu - CLI to simplify adding storages

aravindavk:

### DIFF
--- a/cli/kubectl_kadalu/storage_add.py
+++ b/cli/kubectl_kadalu/storage_add.py
@@ -1,0 +1,143 @@
+import os
+import yaml
+import subprocess
+import tempfile
+import sys
+
+
+KUBECTL_CMD = "kubectl"
+
+
+def storage_add_args(subparsers):
+    parser_add_storage = subparsers.add_parser('storage-add')
+    parser_add_storage.add_argument(
+        "name",
+        help="Storage Name"
+    )
+    parser_add_storage.add_argument(
+        "--type",
+        help="Storage Type",
+        choices=["Replica1", "Replica3"],
+        default="Replica1"
+    )
+    parser_add_storage.add_argument(
+        "--device",
+        help=("Storage device in <node>:<device> format, "
+              "Example: --device kube1.example.com:/dev/vdc"),
+        default=[],
+        action="append"
+    )
+    parser_add_storage.add_argument(
+        "--path",
+        help=("Storage path in <node>:<path> format, "
+              "Example: --path kube1.example.com:/exports/data"),
+        default=[],
+        action="append"
+    )
+    parser_add_storage.add_argument(
+        "--pvc",
+        help="Storage from pvc, Example: --pvc local-pvc-1",
+        default=[],
+        action="append"
+    )
+
+
+def storage_add_validation(args):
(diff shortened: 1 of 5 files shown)


aravindavk:
Not yet tested and pushing it to https://pypi.org/ is pending. Only one sub-command is added now("kubectl kadalu storage-add")

cli/kubectl_kadalu/storage_add.py
```diff
@@ -0,0 +1,122 @@
...
+    if args.pvc:
+        for pvcdata in args.pvc:
+            node = ""
+            pvc = pvcdata
+            if ":" in pvcdata:
+                node, pvc = pvcdata.split(":")
+
+            content["spec"]["storage"].append(
+                {
+                    "node": node,
```
amarts: no need to provide 'node' option for PVC. ideally that gets ignored. Because what happens in k8s is, where the PVC is provisioned, thats where the pod should start.

cli/setup.py
```diff
@@ -0,0 +1,35 @@
...
+    author_email="support@kadalu.io",
+    description="Kadalu Kubectl Plugin",
+    license="Apache-2.0",
+    keywords="kadalu, container, kubernetes, kubectl",
+    url="https://github.com/kadalu/kadalu",
+    long_description="""
+    Kadalu Kubectl Plugin
+    """,
+    classifiers=[
+        "Development Status :: 3 - Alpha",
```
amarts: Looks to be 'Beta' already :-)

cli/kubectl_kadalu/storage_add.py
aravindavk: ack. I will remove this. Added since an example contains node https://github.com/kadalu/kadalu/blob/master/examples/sample-storage-file-device.yaml#L25

cli/setup.py
aravindavk: :)

aravindavk (commented):
Pending:

- Validations
- New "External" options

cli/kubectl_kadalu/storage_add.py
amarts: Yes, realized it... I should fix that and run tests to validate my claim too.

amarts:
> Pending:
> 
> * Validations
> * New "External" options

Do you want to wait for 'External' to get merged? I believe they both can be parallel efforts... and we can leave adding external storage for this tool as an Intern effort (for anyone who wants to contribute new).

aravindavk:
> Do you want to wait for 'External' to get merged? I believe they both can be parallel efforts... and we can leave adding external storage for this tool as an Intern effort (for anyone who wants to contribute new).

Sure

amarts (approved):

amarts:
Meantime, pushing it to pypi should be part of make release, and it should have the same version as our containers.

amarts (approved):

cli/kubectl_kadalu/storage_add.py
amarts: OK, seems like we have some issues when only PVC option is given. Needs more test. When I have both node and PVC, it works fine.

Let this patch consider not providing node when PVC is given. Will fix the PVC example issue separately.

PR #93, opened 2019-12-19
Get kadalu-quotad working as python package

amarts:
This is required step for getting proper support for External
Gluster storage.

More on it @ https://kadalu.io/rfcs/0001-external-gluster-storage.html

This needs suggestion and feedback.

### DIFF
--- a/server/kadalu_quotad/quotad.py
+++ b/server/kadalu_quotad/quotad.py
@@ -9,6 +9,19 @@ import logging
 from kadalulib import execute, logf, CommandException, \
     get_volname_hash, get_volume_path, PV_TYPE_SUBVOL
 
+# Config file for kadalu info
+# Config file format:
+# {
+#    "version": 1,
+#    "bricks": [
+#      "/data/brick1",
+#      "/data/brick2"
+#    ],
+# }
+# We are using a json file, so the same file may get more parameters
+# for other tools
+#
+CONFIG_FILE="/var/lib/glusterd/kadalu.info"
 
 def set_quota(rootdir, subdir_path, quota_value):
     """
@@ -42,8 +55,8 @@ def get_quota_report(rootdir):
                            rootdir=rootdir,
                            err=err.err,
                            ret=err.ret))
-        return []
 
+    return []
 
 def handle_quota(quota_report, brick_path, volname, pvtype):
     """Sets Quota if info file is available"""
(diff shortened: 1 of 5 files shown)


server/setup.py
```diff
@@ -0,0 +1,36 @@
+from setuptools import setup
+
+
+setup(
+    name="kadalu-quotad",
+    version="0.1.0",
+    packages=["kadalu-quotad"],
```
aravindavk: `-` will cause problem for package name. While creating console scripts it imports as `import kadalu-quotad` that may be problem. Change package name and dir name to `kadalu_quotad`

aravindavk:
Current script only crawls one brick. Needs to change this to crawl multiple local bricks

server/kadalu-quotad/quotad.py
```diff
@@ -106,9 +109,20 @@ def start():
...
     while True:
-        crawl()
-        time.sleep(1)
-        if not first_time:
+        brick_path = os.environ["BRICK_PATH"]
+        if brick_path:
+            crawl(brick_path)
+        try:
+            with open(CONFIG_FILE) as f:
+                for brick in f.readlines():
```
amarts: @aravindavk made the crawl based on config file for now here. Isn't this enough?

aravindavk: Sorry I missed this part. This is good

amarts:
I guess this still needs a sysconfig script to start kadalu_quotad. And handle pushing of the package to pypi through make release scripts.

aravindavk:
> I guess this still needs a sysconfig script to start kadalu_quotad. And handle pushing of the package to pypi through make release scripts.

Systemd service file can be included with setup.py packaging itself. Let me search how to include non-code data files.

server/kadalu_quotad/quotad.py
```diff
@@ -6,10 +6,14 @@
 import json
 import logging
 
+# How do we handle this library dependency while packaging ?
```
amarts: @aravindavk @sac need suggestion on how do we handle this issue with pypi package.

aravindavk: For now, we can copy. I will think about it.

aravindavk: I am working on the Makefile changes to create pypi package and push to pypi repo. Let me see how to package without copy

amarts:
I am thinking for the first version, we can just give a package which we can document to run in background, instead of systemd service.

Systemd scripts can come in as a followup.

aravindavk:
> Systemd scripts can come in as a followup.

Sure

server/setup.py
```diff
@@ -0,0 +1,36 @@
...
+
+setup(
+    name="kadalu_quotad",
+    version="0.1.0",
+    packages=["kadalu_quotad"],
+    include_package_data=True,
+    install_requires=[""],
+    entry_points={
+        "console_scripts": [
+            "kadalu_quotad = quotad.start:main"
```
amarts, line 12: I am not sure if this line is valid/alright ? @aravindavk ?

I see that if we get "__main__" then we call start, can I directly map the method here?

aravindavk: `kadalu_quotad.quotad:crawl` if crawl is the starting function when this CLI is called

aravindavk: This can be fixed with the Makefile and Pypi changes.

aravindavk (approved):

PR #94, opened 2019-12-19
Have an unique UID per installation

### LINKED ISSUE #80: Send constant CID (client-id) per deployment
Right now, the analytics we get is skewed as we send cid (ie, client-id) as the timestamp. Ideally it should be a random number (uuid?) constant for deployment.

This can be achieved by having a uuid generated if it doesn't exists in configmap, and while sending analytics, it should be used.

amarts:
Fixes: #80

### DIFF
--- a/lib/kadalulib.py
+++ b/lib/kadalulib.py
@@ -93,7 +93,8 @@ def logging_setup():
     root.addHandler(handler)
 
 
-def send_analytics_tracker(name):
+
+def send_analytics_tracker(name, uid=None):
     """Send setup events to Google analytics"""
     ga_id = "UA-144588868-1" # Static string
     reqheader = {'user-agent': 'Kadalu-App', 'accept': '*/*'}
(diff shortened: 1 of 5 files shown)


manifests/kadalu-operator.yaml
```diff
@@ -191,8 +191,8 @@ spec:
           securityContext:
             capabilities: {}
             privileged: true
-          image: docker.io/kadalu/kadalu-operator:latest
-          imagePullPolicy: Always
+          image: docker.io/kadalu/kadalu-operator:test
```
aravindavk: Is this change only for testing?

lib/kadalulib.py
```diff
@@ -103,8 +104,9 @@ def send_analytics_tracker(name):
     track_title = "Kadalu %s" % name
 
     # TODO: cid should be 1 per installation. How to do it?
```
aravindavk: Todo can be removed?

aravindavk (approved):

manifests/kadalu-operator.yaml
amarts: Yes it was for testing. Removed the file.

PR #97, opened 2019-12-25
Running server/operator/csi processes as PID != 1

### LINKED ISSUE #96: cleanup stuck
While developing, I keep running `./extras/scripts/cleanup` to cleanup kadalu namespace, and restart everything.

Most of the times, I get to below state, after which I have to kill the pod with --force.

```
[amar@localhost kadalu]$ kubectl get pods -nkadalu
NAME                                 READY   STATUS        RESTARTS   AGE
server-storage-pool-1-0-minikube-0   0/2     Terminating   0          65m
```

Would be good to figure out the reason.

aravindavk:
- Introduced a startup script which runs as PID 1 in containers.
  This provides better signals Trapping inside container.
- Bidirectional mount was used between glusterfsd and quotad containers.
  Manual Unmount is required before stopping. Otherwise Kubernetes will
  leave the pod in Terminating state. PreStop script is introduced to send
  SIGTERM to glusterfsd process and unmount the brick.

Fixes: #96

### DIFF
--- a/csi/Dockerfile
+++ b/csi/Dockerfile
@@ -17,9 +17,12 @@ COPY csi/identityserver.py     /kadalu/
 COPY csi/main.py               /kadalu/
 COPY csi/nodeserver.py         /kadalu/
 COPY csi/volumeutils.py        /kadalu/
+COPY lib/startup.sh            /kadalu/startup.sh
 
 COPY templates/Replica1.client.vol.j2 /kadalu/templates/
 COPY templates/Replica3.client.vol.j2 /kadalu/templates/
+
+RUN chmod +x /kadalu/startup.sh
 
 ARG version="(unknown)"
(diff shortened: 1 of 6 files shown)


csi/Dockerfile
```diff
@@ -17,10 +17,13 @@ COPY csi/identityserver.py     /kadalu/
 COPY csi/main.py               /kadalu/
 COPY csi/nodeserver.py         /kadalu/
 COPY csi/volumeutils.py        /kadalu/
+COPY lib/startup.sh            /kadalu/startup.sh
```
amarts, line 20: nit: can be just /kadalu/ in the target like above (for uniformity).

amarts (approved):
Good to merge, we can get minor improvements later.

PR #101, opened 2019-12-27
use sidecar to get gluster mount logs

### LINKED ISSUE #99: Use sidecar container for Logging
CSI Pods mounts Gluster volumes and those logs are not in `stderr` of csi container because these mounts are child processes in CSI pods.

Use another sidecar container which watches on the log files created by these mounts and log to stderr.

amarts:
Fixes: #99

### DIFF
--- a/templates/csi.yaml.j2
+++ b/templates/csi.yaml.j2
@@ -216,6 +216,15 @@ spec:
               mountPath: "/var/lib/gluster"
             - name: gluster-dev
               mountPath: "/dev"
+            - name: varlog
+              mountPath: /var/log/gluster
+        - name: kadalu-logging
+          image: busybox
+          command: ["/bin/sh"]
+          args: ["-c", "while true; do logcnt=$(/bin/ls /var/log/gluster/ | wc -l); if [ ${logcnt} -gt 0 ]; then break; fi; sleep 5; done; tail -f /var/log/gluster/*.log"]
+          volumeMounts:
+            - name: varlog
+              mountPath: "/var/log/gluster"
       volumes:
         - name: plugin-dir
           hostPath:
@@ -235,6 +244,8 @@ spec:
         - name: gluster-dev
           hostPath:
             path: "/dev"
+        - name: varlog
+          emptyDir: {}
 
 ## Deploy CSI Provisioner
 ---
@@ -371,6 +382,15 @@ spec:
               mountPath: "/var/lib/gluster"
             - name: gluster-dev
               mountPath: "/dev"
+            - name: varlog
+              mountPath: "/var/log/gluster"
+        - name: kadalu-logging
+          image: busybox
+          command: ["/bin/sh"]
+          args: ["-c", "while true; do logcnt=$(/bin/ls /var/log/gluster/ | wc -l); if [ ${logcnt} -gt 0 ]; then break; fi; sleep 5; done; tail -f /var/log/gluster/*.log"]
+          volumeMounts:
+            - name: varlog
+              mountPath: "/var/log/gluster"
       volumes:
         - name: socket-dir
           emptyDir:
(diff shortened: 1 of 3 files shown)


aravindavk (approved):
Looks good. `tail -f *.log` prints file name also in the output. That may become lot of logging!

Not an issue now, we can figure that out later.

amarts:
Yes, tried it locally. There is no easy way out with `tail -f`... Mostly need our own logger image, which would output file name (or volumename) per line later.

Also not sure if just mentioning 'image: busybox would work... need to check that too.

amarts:
Ok, sidecar container logic is working to get the logs in `kubectl logs` output :-)

But noticed that there was one more place we were logging to `/dev/stdout` instead of file. Changed that and force-committed. Once that passes test, we can merge this.

aravindavk:
Noticed csi pod restarts, do you see any issues?

```
csi-nodeplugin-zfm4w                 2/3     Error     4          109s

csi-provisioner-0                    3/4     Error     4          109s
```

amarts:
```
csi-nodeplugin-j7fn6                 2/3     CrashLoopBackOff   3          109s
csi-provisioner-0                    3/4     CrashLoopBackOff   3          109s
```

Noticed that the logger fails to start till we setup the config. 

```
tail: can't open '/var/log/gluster/*.log': No such file or directory
tail: no files
```

templates/csi.yaml.j2
```diff
@@ -371,6 +382,15 @@ spec:
               mountPath: "/var/lib/gluster"
             - name: gluster-dev
               mountPath: "/dev"
+            - name: varlog
+              mountPath: "/var/log/gluster"
+        - name: kadalu-logging
+          image: busybox
+          command: ["/bin/sh"]
+          args: ["-c", "while true; do logcnt=$(/bin/ls /var/log/gluster/ | wc -l); if [ \${logcnt} -gt 0 ]; then break; fi; sleep 5; done; tail -f /var/log/gluster/*.log"]
```
amarts: @aravindavk can you think of a better way to do the wait and run? Should we have our own image (busybox + script) instead?

amarts: mainly because this is erroring out with below msg:

```
  File "/kadalu/main.py", line 465, in <module>
    main()
  File "/kadalu/main.py", line 448, in main
    deploy_csi_pods(core_v1_client)
  File "/kadalu/main.py", line 394, in deploy_csi_pods
    execute(KUBECTL_CMD, "create", "-f", filename)
  File "/kadalu/kadalulib.py", line 60, in execute
kadalulib.CommandException: [1] b'clusterrole.rbac.authorization.k8s.io/external-attacher-runner created\nclusterrolebinding.rbac.authorization.k8s.io/csi-attacher-role created\nstatefulset.apps/csi-attacher created\nclusterrole.rbac.authorization.k8s.io/csi-nodeplugin created\nclusterrolebinding.rbac.authorization.k8s.io/csi-nodeplugin created' b'error: error parsing /kadalu/templates/csi.yaml: error converting YAML to JSON: yaml: line 83: found unknown escape character'
```

aravindavk: This looks good to me.

aravindavk: Found this. `-F` option to tail instead of `-f` https://unix.stackexchange.com/questions/31237/how-to-start-tailing-a-file-that-has-not-been-yet-created

aravindavk: Removing `\` before ${logcnt} is showing as valid yaml. Checked online here http://www.yamllint.com/

amarts: I noticed `-F` can't be used in our case.. as we are giving *.log. (-F uses 'name' option, and hence didn't open the file if it was created after the command started).

Noticed that I have to escape twice as its YAML. If I remove the escape, the logcnt was treated as an variable while constructing the string itself. Let me check if \\ solves. Next Iteration, will remove \\ (if this fails).

amarts:
Ok, latest version seems to be all fine.

PR #105, opened 2019-12-27
csi: for now, let all mount logs get into single file

### LINKED ISSUE #103: logging sidecar not picking the latest log file
After merging #101, if we add a new storage to pool, the CSI mount logs are not logged to the stdout.

Expectation is, all of the client logs are in stdout so that can be persisted, and used in debugging without logging into the pods.

amarts:
The issue happened because we use `tail -f *.log` and the `*` is expanded during the initial part of execution, and the files which are available when that gets executed are all written to stdout, but the files which gets created afterwords, they wouldn't.

Hence using same log file for all volumes, which solves the problem, but in a best way.

There is scope for improvement, but can happen from different people :-)

Fixes: #103, #99

### DIFF
--- a/csi/volumeutils.py
+++ b/csi/volumeutils.py
@@ -503,7 +503,8 @@ def mount_glusterfs(volume, target_path):
 
     generate_client_volfile(volume['name'])
     # Fix the log, so we can check it out later
-    log_file = "/var/log/gluster/%s.log" % target_path.replace("/", "-")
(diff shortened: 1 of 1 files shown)


amarts:
#104 tries to solve it properly. If one wants to solve it, they can pick ideas from there.

aravindavk (approved):

Issue #106, opened 2019-12-28
Test: install operator from 0.4 version, and upgrade to latest (any release 0.5/0.6 etc)


amarts:
Ideally, the upgrade should be smooth, without loosing any information about previously created PVs, storage config etc. The process of upgrading the operator as of now is like below:

1. check if operator and other csi drivers are all running.
2. just do `kubectl apply -f new-operator-yaml-file`
3. see that all the new versions should be installed now.

Correct the steps, and document the steps if some other procedures are required for upgrade.

aravindavk:
upgrade testing is needed.

amarts:
no automated tests yet, but the feature is now available with `0.7.6`.

PR #110, opened 2019-12-31
operator: run 'apply' instead of 'create' if the resource exists

amarts:
Updates: #106


amarts:
@aravindavk this is experimental thought.

As we do changes in templates of csi/server pods and storageclass/configmap etc, not updating them if they are present would cause people who installed older versions to not get the proper benefits.

I thought just a `kubectl apply -f ` instead of `kubectl create -f` which would appropriately update the resources. Not sure if thats true.

aravindavk (approved):

PR #115, opened 2020-01-06
cli/kubectl_kadalu: add 'install' subcommand

### LINKED ISSUE #114: enhancement: Add 'install' option to kubectl_kadalu
With [kubectl-kadalu](https://kadalu.io/blog/introducing-kadalu-kubectl-plugin) package, there are possibilities for us to further reduce the 'dependency' of knowing kubernetes for starting kadalu.

One such thing is, removing the need to run `kubectl create -f https://kadalu.io/kadalu-operator.yaml`, instead make it `kubectl kadalu install [--version 1.0.0]` or something similar. That way, all user need to do is `pip3 install kubectl_kadalu` and just use `kubectl kadalu` subcommands to get started.

amarts:
With `install` subcommand, this CLI package now becomes
one tool to install and add storage for kadalu operator.

Invoking it is as simple as `kubectl kadalu install` and
it takes an optional argument of `--version`, which by
default takes latest yaml.

Fixes: #114

### DIFF
--- a/cli/kubectl_kadalu/install.py
+++ b/cli/kubectl_kadalu/install.py
@@ -1,0 +1,55 @@
+import os
+import yaml
+import subprocess
+import tempfile
+import sys
+
+
+KUBECTL_CMD = "kubectl"
+
+
+def install_args(subparsers):
+    parser_install = subparsers.add_parser('install')
+    parser_install.add_argument(
+        "--version",
+        help="Kadalu Version to Install [default: latest]",
+        choices=["0.4.0", "latest"],
+        default="latest"
+    )
+    parser_install.add_argument(
+        "--type",
+        help="Type of installation - k8s/openshift [default: kubernetes]",
+        choices=["openshift", "kubernetes"],
+        default="kubernetes"
+    )
+    return
+
+
+def subcmd_install(args):
+    file_url = "https://raw.githubusercontent.com/kadalu/kadalu/master/manifests"
+    version = ""
+    insttype = ""
+
+    if args.version and args.version != "latest":
+        version = "-%s" % args.version
+
+    if args.type and args.type == "openshift":
+        insttype = "-openshift"
+
+    operator_file = "%s/kadalu-operator%s%s.yaml" % (file_url, insttype, version)
+
+    try:
+        cmd = [KUBECTL_CMD, "apply", "-f", operator_file]
+        resp = subprocess.run(cmd, capture_output=True, check=True,
+                              universal_newlines=True)
+        print("Kadalu operator create request sent successfully")
+        print(resp.stdout)
+        print()
+    except subprocess.CalledProcessError as err:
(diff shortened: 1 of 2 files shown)


cli/kubectl_kadalu/main.py
```diff
@@ -3,13 +3,19 @@
...
+from kubectl_kadalu.install import install_args, subcmd_install
+
 
 def get_args():
     parser = ArgumentParser()
     subparsers = parser.add_subparsers(dest="mode")
 
     storage_add_args(subparsers)
 
+    subparsers = parser.add_subparsers(dest="mode")
```
aravindavk: This line not required again(Line 11 already has it)

cli/kubectl_kadalu/install.py
```diff
@@ -0,0 +1,44 @@
...
+    parser_install.add_argument(
+        "--version",
+        help="Kadalu Version to Install [default: latest]",
+        choices=["0.4.0"],
+        default=None
+    )
+    return
+
+def subcmd_install(args):
+    operator_file = "https://kadalu.io/operator-latest.yaml"
```
aravindavk: How about using file from github repo itself for latest as well. We can avoid copying this file to github.io website on every release

cli/kubectl_kadalu/install.py
```diff
@@ -0,0 +1,44 @@
...
+
+def install_args(subparsers):
+    parser_install = subparsers.add_parser('install')
+    parser_install.add_argument(
+        "--version",
+        help="Kadalu Version to Install [default: latest]",
+        choices=["0.4.0"],
+        default=None
+    )
+    return
```
aravindavk, line 25: one more arg will be good for deployment type: `kubernetes` or `openshift`

cli/kubectl_kadalu/install.py
```diff
@@ -0,0 +1,44 @@
...
+
+def subcmd_install(args):
+    operator_file = "https://kadalu.io/operator-latest.yaml"
+    if args.version:
+        gh_raw_path = "https://raw.githubusercontent.com"
+        repo_path = "kadalu/kadalu"
+        operator_file = "%s/%s/master/manifests/kadalu-operator-%s.yaml" % \
+            (args.version, gh_raw_path, repo_path)
+
```
aravindavk, line 32: Handle deployment type as commented above

cli/kubectl_kadalu/main.py
amarts: Removed

cli/kubectl_kadalu/install.py
amarts: Considering we used it in homepage, it may be required to be done for some more time.

cli/kubectl_kadalu/install.py
amarts: Added.

aravindavk (approved):

PR #116, opened 2020-01-07
Add deletion of CSIDriver

papanito:
Based on the feedback yesterday, it might make sense to add the deletion of the CSIDriver to the cleanup script


amarts:
Looks same as #113. But I like adding of the comment here.

amarts (approved):

PR #134, opened 2020-01-17
travis: ignore few checks in pylint

amarts:
few checks disabled (E0401, E1101):
  * server/glusterfsd.py:34:8: E1101: Module 'xattr' has no 'set' member (no-member)
  * server/glusterfsd.py:35:14: E1101: Module 'xattr' has no 'get' member (no-member)
  * csi/main.py:9:0: E0401: Unable to import 'grpc' (import-error)

Make the test fail in case of warning and errors.

Updates: #63

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -42,6 +42,7 @@ jobs:
   include:
     - name: pylint tests
       script:
+        - pip install requests xxhash grpcio pyxattr kubernetes
         - make pylint || travis_terminate 1;
     - name: kadalu with kube 1.15.0
       script:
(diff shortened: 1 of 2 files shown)


aravindavk:
It looks like good to have these tests. We can ignore a specific case like below.(Add to the respective line)

```
# noqa # pylint: disable=no-member
```

Also installing `grpc` will solve the other error.

aravindavk:
Which `xattr` python lib is installed? `pyxattr` should have that member. Let me check

amarts:
> Which `xattr` python lib is installed? `pyxattr` should have that member. Let me check

I used 'xattr', hence the issue I guess.

aravindavk (approved):

PR #142, opened 2020-01-22
ucli/kubectl_kadalu: add '--external' option

amarts:
With this option, one can add external storage as storage.

also updated doc/external-storage to reflect the possible options.

### DIFF
--- a/cli/kubectl_kadalu/storage_add.py
+++ b/cli/kubectl_kadalu/storage_add.py
@@ -15,8 +15,8 @@ def storage_add_args(subparsers):
     parser_add_storage.add_argument(
         "--type",
         help="Storage Type",
-        choices=["Replica1", "Replica3"],
-        default="Replica1"
+        choices=["Replica1", "Replica3", "External"],
+        default=None
     )
     parser_add_storage.add_argument(
         "--device",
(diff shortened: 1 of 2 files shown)


cli/kubectl_kadalu/storage_add.py
```diff
@@ -15,8 +15,8 @@ def storage_add_args(subparsers):
     parser_add_storage.add_argument(
         "--type",
         help="Storage Type",
-        choices=["Replica1", "Replica3"],
-        default="Replica1"
+        choices=["Replica1", "Replica3", "External"],
+        default=None
```
aravindavk, line 19: If default is `None` then can we make this mandatory. `required=True` makes it mandatory.

cli/kubectl_kadalu/storage_add.py
```diff
@@ -38,17 +38,40 @@ def storage_add_args(subparsers):
         default=[],
         action="append"
     )
+    parser_add_storage.add_argument(
+        "--external",
+        help="Storage from external gluster, Example: --external gluster-node:/gluster-volname",
+        default=[],
+        action="append"
```
aravindavk: If multiple external nodes are not specified then `action="append"` can be removed and `default=""`. (Below it is checked that length == 1)

cli/kubectl_kadalu/storage_add.py
```diff
@@ -38,17 +38,40 @@ def storage_add_args(subparsers):
...
+    num_storages = len(args.device) or len(args.path) or len(args.pvc) or len(args.external)
     if num_storages == 0:
         print("Please specify atleast one storage", file=sys.stderr)
         sys.exit(1)
 
     if (args.type == "Replica1" and num_storages != 1) or (
-            args.type == "Replica3" and num_storages != 3
-    ):
+            args.type == "Replica3" and num_storages != 3) or (
+                args.type == "External" and num_storages != 1
```
aravindavk, line 72: If multiple allowed then change the comparison to `num_storages >= 1`

doc/external-gluster-storage.md
```diff
@@ -1,12 +1,59 @@
...
+
+The best example for this is, what we use in our CI. Checkout this
+[sample yaml file](../examples/sample-external-kadalu-storage.yaml).
+
+Please edit 'gluster_host' and 'gluster_volname' parameters in config section.
+
+You can also setup the external storage through our `kubectl kadalu` CLI.
+
+```console
+kubectl kadalu storage-add store-name --external gluster.kadalu.io:/kadalu
```
aravindavk, line 23: `--type External`

aravindavk:
Also `--kadalu-format` or `--native-format` can be added to support the other type.

cli/kubectl_kadalu/storage_add.py
amarts: No, I don't it mandatory. Idea was if --external is used, I will set this to External, but if there are other things, and no type, then it will be set to Replica1. Because if it picks 'Default', I won't know in --external validation.

cli/kubectl_kadalu/storage_add.py
amarts: Ok! makes sense.

cli/kubectl_kadalu/storage_add.py
amarts: nope, not multiple.

doc/external-gluster-storage.md
amarts: As it was explicitly set, I was thinking of avoiding it in docs.

amarts:
> Also --kadalu-format or --native-format can be added to support the other type.

I feel lets not bother much about `--native-format`. Lets focus only use kadalu format.

cli/kubectl_kadalu/storage_add.py
```diff
@@ -113,6 +135,18 @@ def storage_add_data(args):
...
         return content
 
+    # External details are specified
+    if args.external:
+        for voldata in args.external:
+            node, vol = voldata.split(":")
+            content["spec"]["storage"].append(
+                {
+                    "gluster_host": node,
+                    "gluster_volname": vol
```
aravindavk: Is `/` in volume name ok? If not then use 

```
"gluster_volname": vol.strip("/")
```

aravindavk (approved):

PR #145, opened 2020-01-23
pylint: add kubectl-kadalu CLI to pylint tests

amarts:


aravindavk (approved):

cli/kubectl_kadalu/main.py
```diff
@@ -1,12 +1,15 @@
+"""
+This is an CLI command to handle kadalu executions
+"""
 from argparse import ArgumentParser
 
-from kubectl_kadalu.storage_add import storage_add_args, subcmd_storage_add, \
-    storage_add_validation
+from kubectl_kadalu.storage_add import storage_add_args
```
aravindavk: Removed methods used in Line 30-31. May be we need to skip unused imports

aravindavk (commented):
import related errors need to be ignored in main.py

cli/kubectl_kadalu/main.py
amarts: Good catch! Fixed.

aravindavk (approved):

Issue #45, opened 2019-10-16
RFE: PV resize support (through CSI module)


amarts:
Decide and design possible resize feature.

Will add more details soon. Ping here if you are interested to pick it up

hetii:
I think resizing is one of bit advantage in glusterfs.
As far as I know heketi do it by changing LVM size and apply it in fly, so there is no need to even restart pod that use it. I would love to see similar functionality in kadalu.

aravindavk:
For RWX volumes instead of comparing the Quota already set, compare with the requested quota set size https://github.com/kadalu/kadalu/blob/master/server/quotad.py#L63

```
If new_size > existing_size {
     # Set Quota if free_size_on_disk > (new_size - existing_size)
}

if new_size < existing_size {
     # Set Quota if new_size > du(volume_dir)
}
```

Resizing RWO volumes TBD.

amarts:
For RWO volumes, do something like below (pseudo code):

```
current_size=$(stat -c %s $rwo_filename);
if $new_size_asked > $current_size {
   truncate -s $new_size_asked $rwo_filename;
}

if $new_size_asked < $current_size {
    error_msg;
}
```

And do below on the 'nodeserver' (or wherever we have access to the mount-point)

`xfs_growfs -d $PV_MOUNTPOINT`

aravindavk:
We can split this issue into two separate issues and plan to close RWX resize by v0.6

amarts:
relevant issue! We need pv resize!

aviorma:
any updates?

aravindavk:
> any updates?

Added to v0.8 milestone. Any specific use case in mind? So that it will be useful for our testing.

aviorma:
Very simple use case - Patch and update existing PVC with larger size will extend the glusterfs volume 

StorageClass should has `allowVolumeExpansion: true`
https://kubernetes.io/blog/2018/07/12/resizing-persistent-volumes-using-kubernetes/

schmitch:
currently I stumbled at this project and basically I think kadalu should support the following (btw. I'm already stunned how smooth it works, compared to gluster-kubernetes or heketi):

- resizing the underlying device/path **and** especially the pvc, I think for many applications it's especially useful to actually use kadalu to create a `ReadWriteMany` `pvc` around existing gcloud/aws/azure/etc `pvc`'s, however to be flexible it would be useful to start, with say 100Gib and scale higher if needed (cloud storage is not that cheap).
[...]

PR #152, opened 2020-01-28
addScript

vatsa287:
Script for quick starting the kadalu operator.

### DIFF
--- a/extras/scripts/kadalu-quick-start-generic.sh
+++ b/extras/scripts/kadalu-quick-start-generic.sh
@@ -1,0 +1,107 @@
+#!/bin/bash
+
+#Deploy Kadalu Operator
+echo -e "Do you want to deploy Kadalu or is it already deployed [y/n]"
+read answer1
+if [ $answer1 == 'y' ] ; then 
+
+       sudo kubectl create -f https://kadalu.io/operator-latest.yaml
+       echo "Kadalu Operator Deplyoed"
+
+fi
+#Create Storage Pool to the device, for future claims of this device
+#Using kadalu kubectl plugin
+echo -e "Do you want to create a Storage-Pool [y/n]"
+read answer2
+
+if [ $answer2 == 'y' ] ; then 
+
+	echo -e "Enter the storage-pool name"
+	#example storage-pool-1
+	read storagePool
+	
+	echo -e "Enter the device/host name, for VM use minikube"
+	read device
+	#if using Virtual Machine , minikube
+        
+        echo -e "Do you already have a disk allocation, if yes enter complete path of device or Create a new truncated file [path/create]"
+        read answer3
+        
+        if [ $answer3 == "path" ] ; then
+
+		echo -e "Enter the device file path"
+		read devPath
+
+		#example /home/vatsa/tryKadalu/disk/test-storage
+                sudo kubectl kadalu storage-add $storagePool --device $device:$devPath
+	        echo "Storage Pool with name $storagePool created"
+                
+        else
+
+	        echo -e "Enter the path of the device file to be created"
+	        read dev
+	        
+	        echo -e "Enter the size of the truncated file in Ng or NNNm"
+	        read devSize
+
+	        sudo truncate -s $devSize $dev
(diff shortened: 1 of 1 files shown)


amarts:
@vatsa287 I am not sure such scripts are of a choice to many people because they may have some experience already by the time they start with deploying the project. May be it is a good thing to add in a blog, and say you did use it.

But thinking more about it, you can extend such behavior with `kubectl kadalu` CLI itself. If there are no arguments given, it can ask for next steps interactively. And as it already has options to manage yaml files, that may be a better place to start this.

amarts (changes requested):
marking request changes as I prefer it to be done through CLI at least.

vatsa287:
> marking request changes as I prefer it to be done through CLI at least.

sure thing.

amarts:
IMO, this flow is not required for any k8s user, as most of them would be aware of how to setup the config files! Instead the best PR to work on would be the #45 as a first-timer!

vatsa287:
Yes , I will look into it and try to learn

amarts:
Can you make the script without `kubectl-kadalu` package? but with a template storage yaml file, which gets replaced with proper variables based on the arguments?

extras/scripts/kadalu-quick-start-generic.sh
```diff
@@ -0,0 +1,107 @@
+#!/bin/bash
+
+#Deploy Kadalu Operator
+echo -e "Do you want to deploy Kadalu or is it already deployed [y/n]"
+read answer1
+if [ $answer1 == 'y' ] ; then 
+
+       sudo kubectl create -f https://kadalu.io/operator-latest.yaml
```
amarts, line 8: sudo kubectl apply -f https://raw.githubusercontent.com/kadalu/kadalu/devel/manifests/kadalu-operator.yaml

extras/scripts/kadalu-quick-start-generic.sh
```diff
@@ -0,0 +1,107 @@
...
+        echo -e "Do you already have a disk allocation, if yes enter complete path of device or Create a new truncated file [path/create]"
+        read answer3
+        
+        if [ $answer3 == "path" ] ; then
+
+		echo -e "Enter the device file path"
+		read devPath
+
+		#example /home/vatsa/tryKadalu/disk/test-storage
+                sudo kubectl kadalu storage-add $storagePool --device $device:$devPath
```
amarts, line 36: no need for kubectl kadalu... the script should run without python dependency!

extras/scripts/kadalu-quick-start-generic.sh
```diff
@@ -0,0 +1,107 @@
...
+        cp ../../examples/sample-app.yaml /tmp/${podFileName}.yaml
+	sed -i s/pod1/$podName/  /tmp/${podFileName}.yaml
+	sed -i s/pv1/$pvcName/  /tmp/${podFileName}.yaml
+	sudo kubectl create -f /tmp/$podFileName.yaml
+        
+        echo "Sleep 100s for status update"
+        sleep 100
+ 
+        #Status of the Pod
+        sudo kubectl get pods
```
amarts, line 100: can you use `getopt` and take these variables as arguments instead. Also provide '--help'.

extras/scripts/kadalu-quick-start-generic.sh
```diff
@@ -0,0 +1,107 @@
...
+	read podName
+
+
+        cp ../../examples/sample-app.yaml /tmp/${podFileName}.yaml
+	sed -i s/pod1/$podName/  /tmp/${podFileName}.yaml
+	sed -i s/pv1/$pvcName/  /tmp/${podFileName}.yaml
+	sudo kubectl create -f /tmp/$podFileName.yaml
+        
+        echo "Sleep 100s for status update"
+        sleep 100
```
amarts, line 97: no need to sleep 100, but do it like https://github.com/kadalu/kadalu/blob/devel/tests/minikube.sh#L11

amarts (changes requested):

vatsa287:
Changes done through patch #327 .

PR #166, opened 2020-01-31
Minor quick start fix

amarts:


amarts:
Considering this would impact every user, merging immediately.

PR #160, opened 2020-01-29
Thin arbiter

amarts:
Supported the thin-arbiter volume type as per the RFC specified @

https://kadalu.io/rfcs/0003-kadalu-thin-arbiter-support

Provided option to take default tie-breaker node if no tie-breaker is provided.

Fixes: #37

PR #168, opened 2020-02-04
Add details to run a tie-breaker container-image

amarts:
Updates #160

### DIFF
--- a/extras/tie-breaker/tie-breaker.yaml
+++ b/extras/tie-breaker/tie-breaker.yaml
@@ -1,0 +1,20 @@
+# -*- mode: yaml -*-
+---
+apiVersion: v1
+kind: Pod
+metadata:
+  name: tie-breaker-pod
+  labels:
+    app: gluster-tie-breaker
+spec:
+  containers:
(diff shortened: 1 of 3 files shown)


extras/tie-breaker/tie-breaker.yaml
```diff
@@ -0,0 +1,20 @@
...
+apiVersion: v1
+kind: Pod
+metadata:
+  name: tie-breaker-pod
+  labels:
+    app: gluster-tie-breaker
+spec:
+  containers:
+  - name: tie-breaker
+    image: docker.io/kadalu/gluster-tie-breaker:latest
```
aravindavk: name mentioned above is "kadalu-tiebreaker-app" (https://github.com/kadalu/kadalu/pull/168/files#diff-bacb243d4cac3bc29dd35ad5d1ac8469R16)

amarts: not sure what I was thinking. Good catch! Fixed in the update.

aravindavk (approved):
LGTM

PR #170, opened 2020-02-05
requirements.txt

### LINKED ISSUE #60: Install project  dependencies
This project has dependencies it would be good to add those into a `requirement.txt` file and update Readme to install same to setup development environment

vatsa287:
Fixes: #60

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -3,4 +3,4 @@ xxhash
 grpcio
 pyxattr
 kubernetes
-#any feedbacks on version_no?
+
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,0 +1,6 @@
+requests
+xxhash
+grpcio
+pyxattr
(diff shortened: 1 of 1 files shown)


requirements.txt
```diff
@@ -0,0 +1,6 @@
+requests
+xxhash
+grpcio
+pyxattr
+kubernetes
+#any feedbacks on version_no?
```
amarts: No need to have this here.. it should be a comment while reviewing. Remove this.

amarts (commented):
Also, commit msg can have 'Fixes: #60' instead of 'Issue #60'.

requirements.txt
vatsa287: Updated

amarts (commented):
/lgtm

aravindavk (approved):

Issue #37, opened 2019-10-01
RFE: thin-arbiter (metro cluster) support


amarts:
GlusterFS-7 is releasing with thin-arbiter support. Once released, if kadalu supports thin-arbiter cluster support, it will help many users who have only 2 data centers.

PR #171, opened 2020-02-07
extras/tie-breaker: fix the starting errors

amarts:
With this patch, once the images are pused to docker.io, we can get
thin-arbiter running on any machine like below:

`docker run --privileged -p 24007:24007 -v ${hostbrickdir}:/mnt/brick docker.io/kadalu/kadalu-tiebreaker:latest`

We expect this to be the easiest way of running a thin-arbiter on
any machine for our users, along with other apps too.

Updates: #37


amarts:
[skip ci]
skipping CI as this is not touching any files related to kadalu

aravindavk (approved):

PR #174, opened 2020-02-10
csi/controller: use only subdir model for all PVs

amarts:
Ideally, both RWX and RWO mode we provide in default mode, are part of `volumeMode: Filesystem` in PV Claim.

So, with this patch, we are keeping expectation same. PV volumes come as filesystem.

Good part of this is, a single PV can be of minimum size of 1Mi (1 MB) min. A great usecase to testing scale, where we can use 1000 PVs with just 1GB backend storage.


aravindavk (approved):

amarts:
This needs to go through a perf testing before v1.0, but I feel a good change to get in.

PR #179, opened 2020-02-11
storage class initializing and check sanitized

amarts:
* storageclass gets initialized only after storage options are available for kadalu.

* in CSI if a request reaches controller, and is not of kadalu type, it would be denied PV create.


amarts:
Discussed with @aravindavk, and we agreed, we should have 'kadalu' as default if there are no other default storageclass in the system. So, it is not a bad option. Need to fix it differently.

amarts:
Not required for now. We will see the behavior.

PR #181, opened 2020-02-12
Liveness and Readiness probe for Glusterfsd container

aravindavk:
Added Liveness and Readiness probe to see glusterfsd
container is started and port 24007 is ready to use.

Also check Glusterfs mounted or not using `ps -ax | grep glusterfs`
instead of using `os.path.ismount` command.

### DIFF
--- a/csi/volumeutils.py
+++ b/csi/volumeutils.py
@@ -13,7 +13,7 @@ from jinja2 import Template
 
 from kadalulib import execute, PV_TYPE_SUBVOL, PV_TYPE_VIRTBLOCK, \
     get_volname_hash, get_volume_path, logf, makedirs, CommandException, \
-    retry_errors
+    retry_errors, is_gluster_mount_proc_running
 
 
 GLUSTERFS_CMD = "/usr/sbin/glusterfs"
@@ -473,6 +473,13 @@ def mount_volume(pvpath, target_path, pvtype, fstype=None):
     os.chmod(target_path, 0o777)
 
 
+def unmount_glusterfs(target_path):
+    """Unmount GlusterFS mount"""
+    volname = os.path.basename(target_path)
+    if is_gluster_mount_proc_running(volname):
+        execute(UNMOUNT_CMD, target_path)
+
+
 def unmount_volume(target_path):
     """Unmount a Volume"""
     if os.path.ismount(target_path):
@@ -503,11 +510,13 @@ def generate_client_volfile(volname):
 
 def mount_glusterfs(volume, target_path):
     """Mount Glusterfs Volume"""
-    if not os.path.exists(target_path):
-        makedirs(target_path)
+    if volume["type"] == "External":
+        volname = volume['g_volname']
+    else:
+        volname = volume["name"]
 
-    # Ignore if already mounted
-    if os.path.ismount(target_path):
+    # Ignore if already glusterfs process running for that volume
+    if is_gluster_mount_proc_running(volname):
         logging.debug(logf(
             "Already mounted",
             mount=target_path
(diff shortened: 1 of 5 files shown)


aravindavk:
Testing pending. I will update the issue once testing is complete.

aravindavk:
First error when Volume mount was successful but statvfs failed since brick was not up

```
[2020-02-25 15:26:02,171] DEBUG [controllerserver - 165:CreateVolume] - Got list of hosting Volumes      volumes=storage-pool-1
  File "/kadalu/controllerserver.py", line 170, in CreateVolume
    hostvol = mount_and_select_hosting_volume(host_volumes, pvsize)
  File "/kadalu/volumeutils.py", line 156, in mount_and_select_hosting_volume
    mntdir_stat = os.statvfs(mntdir)
OSError: [Errno 107] Transport endpoint is not connected: '/mnt/storage-pool-1'
```

After this mount becomes stale? Looks like `os.ismount` will not show this as mounted.

Mount will be retried for every call after, but mount command fails(May be for existing stale mount process?)

```
[2020-02-25 15:26:02,253] ERROR [volumeutils - 528:mount_glusterfs] - error to execute command   volume={'name': 'storage-pool-1', 'type': 'Replica1', 'g_volname': None, 'g_host': None, 'g_options': None} cmd=['/usr/sbin/glusterfs', '--process-name', 'fuse', '-l', '/var/log/gluster/gluster.log', '--volfile-id', 'storage-pool-1', '-f', '/kadalu/volfiles/storage-pool-1.client.vol', '/mnt/storage-pool-1'] error=[1] b'' b''
[2020-02-25 15:26:02,253] ERROR [_server - 445:_call_behavior] - Exception calling application: [1] b'' b''
  File "/usr/local/lib64/python3.7/site-packages/grpc/_server.py", line 435, in _call_behavior
  File "/kadalu/controllerserver.py", line 170, in CreateVolume
    hostvol = mount_and_select_hosting_volume(host_volumes, pvsize)
  File "/kadalu/volumeutils.py", line 150, in mount_and_select_hosting_volume
    mount_glusterfs(volume, mntdir)
  File "/kadalu/volumeutils.py", line 530, in mount_glusterfs
    raise err
  File "/kadalu/volumeutils.py", line 522, in mount_glusterfs
    execute(*cmd)
  File "/kadalu/kadalulib.py", line 61, in execute
kadalulib.CommandException: [1] b'' b''
```

Fixes required are:

- If mount process re-connects automatically to bricks when they are available, then handle `ENOTCONN` and add retry to all commands executed via mount(`statvfs`, `mkdir`, etc)
- Use other method to decide mount exists and then remount
- Use unmount force even if `os.path.ismount` says no mounts exist(Ignore `umount` error if mount not exists error)

csi/volumeutils.py
```diff
@@ -533,6 +533,12 @@ def mount_glusterfs(volume, target_path):
         return
 
     with mount_lock:
+        # Unmount to make sure no previous stale mounts
+        try:
+            execute(UNMOUNT_CMD, target_path)
```
amarts: does it 'umount()' an active volume? i see mount_glusterfs() gets called from multiple places.

templates/csi.yaml.j2
```diff
@@ -232,7 +232,7 @@ spec:
           args:
             - "--provisioner=kadalu"
             - "--csi-address=$(ADDRESS)"
-            - "--connection-timeout=15s"
+            - "--connection-timeout=2m"
```
amarts: Should it be few seconds more? I see 'initialDelay' itself is 120 seconds. I think we may have to give some delta (at least of 3-7 seconds) between these 2?

csi/volumeutils.py
aravindavk: coming here after `ismount` check. So should be fine I guess. But this I added before checking `ismount` not detecting stale mounts. We need to think about a different solution instead of using `ismount`

templates/csi.yaml.j2
aravindavk: Ok, I will do that

aravindavk:
@amarts changed the method to check mount process exists.

amarts (approved):

PR #193, opened 2020-02-26
csi: make sure we never overwrite

amarts:
There are reports of possible PV create request repeat,
which gets error like below:

```
b'mkfs.xfs: /mnt/snowstorm/virtblock/ed/d3/pvc-7e309eed-7654-4aee-a014-a3c48b712756 appears to contain an existing filesystem (xfs).
```

This patch tries to make sure we will never have a case where there would be PV which is already created being created as volume again.

### DIFF
--- a/csi/volumeutils.py
+++ b/csi/volumeutils.py
@@ -207,6 +207,14 @@ def create_virtblock_volume(hostvol_mnt, volname, size):
         path=os.path.dirname(volpath)
     ))
 
+    if os.path.exists(volpath_full):
+        rand = time.time()
+        logging.info(logf(
+            "Getting 'Create request' on existing file, renaming.",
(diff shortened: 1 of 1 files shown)


csi/volumeutils.py
```diff
@@ -207,6 +208,14 @@ def create_virtblock_volume(hostvol_mnt, volname, size):
         path=os.path.dirname(volpath)
     ))
 
+    if os.path.exists(volpath_full):
+        rand = uuid.uuid1()
```
aravindavk: I think adding `int(time.time())` is good compared to random UUID. It helps to identify the volume based on the time when it was rolled over.

amarts: Done

aravindavk (approved):

PR #194, opened 2020-02-26
Release workflow fix

amarts:
* Add dependency of python, and related packages.


amarts:
Tested these separately in my repo.

https://github.com/amarts/kadalu/runs/470307477?check_suite_focus=true

Now, there are further enhancement
* ignore tagging of 'rc/pre' etc in tag name to be pushed as 'latest'. 
* do nightly (or after merge) push to 'master' tag
* pypi shows a release as pre-release because the version has `rc` in it.

PR #195, opened 2020-02-27
tests: add tests for kubectl_kadalu

amarts:


aravindavk (approved):

amarts:
>  Warning  FailedScheduling  18s (x7 over 3m26s)  default-scheduler  0/1 nodes are available: 1 node(s) didn't match node selector.

aravindavk:
> Warning FailedScheduling 18s (x7 over 3m26s) default-scheduler 0/1 nodes are available: 1 node(s) didn't match node selector.

May be issue with latest minikube. In my local setup, I observed that the node name(`kubectl get nodes`) was used from `hostname` of host instead of `minikube`.  Set the `hostname minikube` before running the tests.

templates/server.yaml.j2
```diff
@@ -46,8 +46,8 @@ spec:
           readinessProbe:
             tcpSocket:
               port: 24007
-            initialDelaySeconds: 120
-            periodSeconds: 10
+            initialDelaySeconds: 7
+            periodSeconds: 3
```
amarts: Looks like this can be kept at 10. No need to reduce this.

tests/kubectl_kadalu_tests.sh
```diff
@@ -0,0 +1,53 @@
...
+
+    kubectl kadalu install --local-yaml manifests/kadalu-operator.yaml
+}
+
+function test_storage_add() {
+    sed -i -e "s/DISK/${DISK}/g" tests/get-minikube-pvc.yaml
+    kubectl create -f tests/get-minikube-pvc.yaml
+
+    sleep 1
+    kubectl kadalu storage-add test-volume3 --type Replica3 --device minikube:/mnt/${DISK}/file3.1 --path minikube:/mnt/${DISK}/dir3.2 --pvc local-pvc
```
amarts: looks like each of these commands failing should result in failure of the test. Please fix.

amarts:
Finally, we are ready for this. Will address the above comments in next push!

tests/kubectl_kadalu_tests.sh
amarts: Done!

templates/server.yaml.j2
amarts: Done

amarts:
Merging as this has multiple fixes, and discussed and reviewed the flow with @aravindavk already on this.

PR #210, opened 2020-03-05
csi: fix external volume mount issues

### LINKED ISSUE #209: Deleting external volume and then creating a new one fails
I've successfully been able to set up external storage using kubectl-kadalu (v0.6.0) and then a PVC thanks to your help on the slack channel. I then successfully used that PVC in a test deployment. This all worked fine.

Next, I deleted the PVC and implemented kadalu-quotad as a service on my GlusterFS nodes. I then wanted to provision a new PVC using the same storage class that was previously set up when I added the storage using kubectl-kadalu.

This PVC goes into "pending", however. Using the same storage class for other workloads (Prometheus and Grafana) gives the same results.

Please advise how to fix this, thank you!

I've attached the csi-provisioner-0 log.
[csi-provisioner-0.log](https://github.com/kadalu/kadalu/files/4286502/csi-provisioner-0.log)

amarts:
We need to clearly have details on how the mount point looks
for each PV (and storage host volumes). This consistency will
help us to have same consistent behavior.

Fixes: #209

### DIFF
--- a/lib/kadalulib.py
+++ b/lib/kadalulib.py
@@ -42,15 +42,15 @@ def retry_errors(func, args, errors, timeout=130, interval=2):
             raise
 
 
-def is_gluster_mount_proc_running(volname, target_path):
+def is_gluster_mount_proc_running(volname, mountpoint):
     """
     Check if glusterfs process is running for the given Volume name
     to confirm Glusterfs process is mounted
     """
     cmd = (
         r'ps ax | grep -w "/usr/sbin/glusterfs" '
-        r'| grep "\-\-volfile\-id %s" '
-        r'| grep -q "%s"' % (volname, target_path)
+        r'| grep -w "\-\-volfile\-id %s" '
+        r'| grep -w -q "%s"' % (volname, mountpoint)
     )
 
     proc = subprocess.Popen(cmd, shell=True, stderr=None,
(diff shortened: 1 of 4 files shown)


StoneCut:
I manually applied the above patch to "kadalulib.py" and killed glusterfs. This fixed the issue for me. Thank you again!

lib/kadalulib.py
```diff
@@ -47,10 +47,11 @@ def is_gluster_mount_proc_running(volname, target_path):
     Check if glusterfs process is running for the given Volume name
     to confirm Glusterfs process is mounted
     """
+    # TODO: see if you can use '-w' instead
     cmd = (
         r'ps ax | grep -w "/usr/sbin/glusterfs" '
-        r'| grep "\-\-volfile\-id %s" '
-        r'| grep -q "%s"' % (volname, target_path)
+        r'| grep "\-\-volfile\-id %s " '
+        r'| grep -q " %s"' % (volname, target_path)
```
aravindavk: how space before helps? We can use `-w` or `egrep -q "%s$" % target_path`

aravindavk: Also can be changed to `mountpath`

amarts: Done!

PR #225, opened 2020-04-28
Improvement to tests

amarts:
* base-image: fedora31 has latest glusterfs by default
* travis: install conntrack (required for 1.18.x onwards)
* tests: add a job for 1.18.2 version (remove 1.16.x)
* tests: add hostname instead of 'minikube'
* change minikube version to 1.7.2 (latest is 1.9.2)

Signed-off-by: Amar Tumballi <amar@kadalu.io>

PR #226, opened 2020-04-28
Node affinity added to Storage filter

aravindavk:
This new filter is to support Local PV usecase. If multiple
Replica 1 storage pools available in cluster then create a
Storage class for required node affinity. If a PV is claimed
based on this storage class then it behaves just like Local storage
but with dynamic provisioning.

Example storage class:

```
kind: StorageClass
apiVersion: storage.k8s.io/v1
metadata:
  name: kadalu.replica1.node1
provisioner: kadalu
parameters:
  node_affinity: "node1"  # As shown in `kubectl get nodes`
```

### DIFF
--- a/csi/volumeutils.py
+++ b/csi/volumeutils.py
@@ -56,48 +56,116 @@ class Volume():
         return self.volname
 
 
-def get_pv_hosting_volumes(filters=None):
+def filter_node_affinity(volume, filters):
+    """
+    Filter volume based on node affinity provided
+    """
+    node_name = filters.get("node_affinity", None)
+    if node_name is not None:
+        # Node affinity is only applicable for Replica1 Volumes
+        if volume["type"] != "Replica1":
+            return None
+
+        # Volume is not from the requested node
+        if node_name != volume["bricks"][0]["kube_hostname"]:
+            return None
+
+    return volume
+
+
+def filter_storage_name(volume, filters):
+    """
+    filter volume based on the name provided in filter
+    """
+    storage_name = filters.get("storage_name", None)
+    if storage_name is not None and storage_name != volume["volname"]:
+        return None
+
+    return volume
+
+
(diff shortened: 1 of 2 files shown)


aravindavk:
It looks like Travis tests not triggered. @amarts any clue?

amarts (approved):

amarts:
Checking... Trying with #225 right now.. there I see Travis build triggerring!

amarts:
@aravindavk one possibility! Can you commit to your repo/fork, and open a PR ? Just a small guess that because the changes are pushed to another branch in 'origin' it would have not triggered!

aravindavk:
oh ok, let me see. I opened PR in the same repo but a different branch

aravindavk:
refreshed on the same branch. Now I see Travis triggered.

amarts:
interesting glitch !

amarts:
I suspect that the tests would fail anyways. Looks like we need to get #225 merged regardless.

amarts:
Not sure if this is still critical: 

failed to provision volume with StorageClass "kadalu.replica3": rpc error: code = Unknown desc = Exception calling application: local variable 'volume' referenced before assignment

csi/volumeutils.py
```diff
@@ -56,48 +56,116 @@ def get(self):
...
+            filtered_data = True
+            for filter_func in filter_funcs:
+                filtered = filter_func(data, filters)
+                # Node affinity is not matching for this Volume,
+                # Try other volumes
+                if filtered is None:
+                    filtered_data = False
+                    logging.debug(logf(
+                        "Volume doesn't match the filter",
+                        volname=volume["name"],
```
amarts: volume here wouldn't be defined yet.

aravindavk: fixed. Wondering why pylint didn't picked this..

PR #235, opened 2020-05-02
github: add buildx

amarts:
* Currently added another workflow other than already existing 'Release' specific `dockerpush.yml`
* Plan is to run a buildx for every master push, once this is successful, so people can have official images for testing newer features!
* `xxhash` pip install had issues on arm64 architecture.. for now, replaced that with std hashlib. Need to investigate further

* Noticed we can't get 'secrets' used in github actions for `pull_requests` job! only available for operations on master branch like push or tags

* buildx needed to have 'name' field in --output to work properly, wasted more than few hours because ot that!

* fedora:31 images are not present for linux/686, arm/v6, arm/v7, so not having it in build.

[skip ci]


amarts:
Overrides #230

amarts:
Finally was able to get `pip install grpcio`  pass... only major issue is, it took ~1hr 20mins to get compiled and installed. Will revert to proper steps and all arch now, and trigger a run, it may take 4hr+ to complete, but would be good to have master tag working on all arch!

amarts:
Please merge with 'Squash and Merge' .... Some tests are good... people can help testing with `kubectl create -f https://raw.githubusercontent.com/kadalu/kadalu/49e304563ed50e76969e77ed4ec8dc568d2fd977/manifests/kadalu-operator-master.yml` on arm64 architecture

amarts:
notice that s390x / ppc64le still threw some issue related to grpcio... So, the first cut only has support for amd64/arm64

server/setup.py
```diff
@@ -11,7 +11,7 @@ def version():
     version=version(),
     packages=["kadalu_quotad"],
     include_package_data=True,
-    install_requires=["requests", "xxhash"],
+    install_requires=["requests", "cython", "xxhash"],
```
amarts: Not required! remove

requirements.txt
```diff
@@ -1,8 +1,9 @@
 requests
-xxhash
-grpcio
-pyxattr
 kubernetes
 jinja2
 setuptools
 twine
+cython
```
amarts: not required!

server/setup.py
amarts: Removed

requirements.txt
amarts: Done

aravindavk (approved):

amarts:
If you want to try now, just run `kubectl create -f https://raw.githubusercontent.com/kadalu/kadalu/master/manifests/kadalu-operator-master.yml` on ARM64 machines

PR #259, opened 2020-05-24
README: Add Pronounce option

amarts:


amarts:
I have added link to 'raw' file content from the 'master' branch, so it can be used only after the merge!

aravindavk (approved):

PR #263, opened 2020-05-26
Ubuntu focal update as base image

amarts:
[WIP]


amarts:
Needs to be opened against `devel` branch now.

PR #262, opened 2020-05-26
CSI proto file update and csi sidecar containers image update

aravindavk:
- CSI proto file updated to v1.2.0 and regenerated grpc python
  code by running `make build-grpc`
- CSI provisioner container upgraded to v1.6.0
- CSI attacher container upgraded to v2.2.0
- CSI node registrar container upgraded to v1.3.0

Signed-off-by: Aravinda Vishwanathapura <aravinda@kadalu.io>

PR #273, opened 2020-07-02
[Draft] csi: use v2.2.0 instead of v1.0.1

amarts:
Trying to do this to understand if the newer versions support multiarch!


amarts:
#262 does it better!

PR #279, opened 2020-07-20
csi/controller: check for existence of hostvoltype

### LINKED ISSUE #277: hostvoltype might be NONE
When creating your own storageclass the hostvoltype might not be set.

Normally kadalu storage classes have the parameter `hostvol_type`:
`kubectl get storageclasses.storage.k8s.io kadalu.replica1 -o yaml`
```yaml
apiVersion: storage.k8s.io/v1
kind: StorageClass
metadata:
  creationTimestamp: "2020-07-07T12:20:03Z"
  name: kadalu.replica1
  resourceVersion: "11672"
  selfLink: /apis/storage.k8s.io/v1/storageclasses/kadalu.replica1
  uid: 7761a3a0-14fd-4600-83ca-b98d333a8738
parameters:
  hostvol_type: Replica1
provisioner: kadalu
reclaimPolicy: Delete
volumeBindingMode: Immediate
```

However if we create our own storageClass:
```yaml
kind: StorageClass
apiVersion: storage.k8s.io/v1
metadata:
  name: kadalu.fast-storage
provisioner: kadalu
parameters:
```
This parameter is missing:
`kubectl get storageclasses.storage.k8s.io kadalu.fast-storage -o yaml`
```yaml
apiVersion: storage.k8s.io/v1
kind: StorageClass
metadata:
  creationTimestamp: "2020-07-11T19:33:43Z"
  name: kadalu.fast-storage
  resourceVersion: "1358515"
  selfLink: /apis/storage.k8s.io/v1/storageclasses/kadalu.fast-storage
  uid: e2c9becf-da44-4bbc-9585-344fa480a612
parameters:
provisioner: kadalu
reclaimPolicy: Delete
volumeBindingMode: Immediate
```

The createVolumeReponse then fails.

https://github.com/kadalu/kadalu/blob/274934b3c102391ac2a9a61059e77194d93926ab/csi/controllerserver.py#L205

amarts:
Fixes: #277

### DIFF
--- a/csi/controllerserver.py
+++ b/csi/controllerserver.py
@@ -26,7 +26,7 @@ class ControllerServer(csi_pb2_grpc.ControllerServicer):
     volume mount and PV creation.
     Ref:https://github.com/container-storage-interface/spec/blob/master/spec.md
     """
-    # noqa # pylint: disable=too-many-locals,too-many-statements
+    # noqa # pylint: disable=too-many-locals,too-many-statements,too-many-branches
     def CreateVolume(self, request, context):
         start_time = time.time()
         logging.debug(logf(
@@ -199,6 +199,9 @@ class ControllerServer(csi_pb2_grpc.ControllerServicer):
         ))
 
         update_free_size(hostvol, request.name, -pvsize)
+
+        if not hostvoltype:
+            hostvoltype = "unknown"
 
         send_analytics_tracker("pvc-%s" % hostvoltype, uid)
         return csi_pb2.CreateVolumeResponse(


aravindavk:
What is the expected behaviour when hostvol_type is unknown? Don't select any volume or select from any volume type?

If we need second behaviour then fix required in this function as well. https://github.com/kadalu/kadalu/blob/master/csi/volumeutils.py#L88

close2:
> What is the expected behaviour when hostvol_type is unknown? Don't select any volume or select from any volume type?

Not knowing the `hostvol_type` at this point shouldn't matter.
A suitable brick/volume has already been selected.
It's the `CreateVolumeResponse` which fails:
```python
        return csi_pb2.CreateVolumeResponse(
            volume={
                "volume_id": request.name,
                "capacity_bytes": pvsize,
                "volume_context": {
                    "type": hostvoltype,
                    "hostvol": hostvol,
                    "pvtype": pvtype,
                    "path": vol.volpath,
                    "fstype": "xfs"
                }
            }
        )
```
Which tries to fill `volume_context` of the Protobuffer and fails:
```
  // Opaque static properties of the volume. SP MAY use this field to
  // ensure subsequent volume validation and publishing calls have
  // contextual information.
  // The contents of this field SHALL be opaque to a CO.
  // The contents of this field SHALL NOT be mutable.
  // The contents of this field SHALL be safe for the CO to cache.
  // The contents of this field SHOULD NOT contain sensitive
  // information.
  // The contents of this field SHOULD NOT be used for uniquely
  // identifying a volume. The `volume_id` alone SHOULD be sufficient to
  // identify the volume.
  // A volume uniquely identified by `volume_id` SHALL always report the
  // same volume_context.
  // This field is OPTIONAL and when present MUST be passed to volume
  // validation and publishing calls.
  map<string, string> volume_context = 3;
```

The comment indicates that the `volume_context` itself is already optional.
Unless this comment is wrong, it shouldn't matter if we return the actual `hostvol_type`.

The `hostvol_type` could also be extracted from the brick/volume itself.  But that's probably not necessary.

aravindavk:
Thanks @close2 for the clarification.

aravindavk (approved):

PR #285, opened 2020-08-24
Replica3: remove perf xlators from client volfile

amarts:
Also remove all the default value of volfile, and keep it minimal


amarts:
This is done as there were some application errors with current volume. By starting with minimal options, it becomes easy to identify the issue.

amarts:
Closing to have it in another PR.

Issue #287, opened 2020-08-25
problem running elasticsearch cluster on replica3


davehatton:
I have an elasticsearch cluster deployed from a statefulset  that is working when using heketi/gluster but fails to start when migrated to kadalu/replica3 storage.

The key message from the application appears to be 

```
2020-08-25T08:46:10.799184289Z java.lang.IllegalStateException: failed to obtain node locks, tried [[/usr/share/elasticsearch/data]] with lock id [0]; maybe these locations are not writable or multiple nodes were started without increasing [node.max_local_storage_nodes] (was [1])?
2020-08-25T08:46:10.799197336Z Likely root cause: java.io.IOException: Value too large for defined data type
```

My storage is presented using 

```
---
 kind: KadaluStorage
 apiVersion: kadalu-operator.storage/v1alpha1
 metadata:
   name: storage-pool-1
 spec:
   type: Replica3
   storage:
   - node: k807
     device: /dev/sdb
   - node: k808
     device: /dev/sdb
   - node: k809
     device: /dev/sdb

...
```

and my stateful set is:

```
---
 apiVersion: apps/v1
 kind: StatefulSet
 metadata:
   namespace: monitoring
   name: es-cluster
 spec:
   serviceName: elasticsearch
   replicas: 3
   selector:
     matchLabels:
       app: elasticsearch
   template:
     metadata:
       labels:
         app: elasticsearch
     spec:
       containers:
       - name: elasticsearch
         image: docker.elastic.co/elasticsearch/elasticsearch:7.8.0
         resources:
           limits:
             cpu: 1000m
           requests:
             cpu: 100m
         ports:
         - containerPort: 9200
           name: rest
           protocol: TCP
         - containerPort: 9300
           name: inter-node
           protocol: TCP
         volumeMounts:
         - name: data
           mountPath: /usr/share/elasticsearch/data
         env:
           - name: cluster.name
             value: k8s-logs
           - name: node.name
             valueFrom:
               fieldRef:
                 fieldPath: metadata.name
           - name: discovery.seed_hosts
             value: "es-cluster-0.elasticsearch,es-cluster-1.elasticsearch,es-cluster-2.elasticsearch"
           - name: cluster.initial_master_nodes
             value: "es-cluster-0,es-cluster-1,es-cluster-2"
           - name: ES_JAVA_OPTS
             value: "-Xms512m -Xmx512m"
       initContainers:
       - name: fix-permissions
         image: busybox
[...]

PR #288, opened 2020-08-25
Tests: add a test for elastic-search statefulset

amarts:
Reference: #287 


amarts:
Have to rebase to 'devel'

PR #291, opened 2020-09-02
Replica3: remove perf xlators from client volfile

amarts:
Also remove all the default value of volfile, and keep it minimal


amarts:
Previously was @ #285 (but as we use 'buildx_test' branch to get the master tag tested, created another PR).

amarts:
Opening again on top of `devel` branch.

PR #293, opened 2020-09-02
Ubuntu focal

amarts:


amarts:
Similar to #263 but now opened on `devel` branch.

amarts:
#385 does this better!

PR #298, opened 2020-09-04
when using pvc's it isn't needed to mount /usr/lib/modules

schmitch:
currently this fixes a bug in some distributions that can't
mount /usr/lib/modules because it does not exist and
/usr is read-only (GKE)

### DIFF
--- a/templates/server.yaml.j2
+++ b/templates/server.yaml.j2
@@ -85,9 +85,11 @@ spec:
             - name: gluster-cgroup
               mountPath: "/sys/fs/cgroup"
               readOnly: true
+{%- if pvc_name == "" %}
             - name: gluster-kmods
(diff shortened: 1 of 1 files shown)


templates/server.yaml.j2
```diff
@@ -85,9 +85,11 @@ spec:
             - name: gluster-cgroup
               mountPath: "/sys/fs/cgroup"
               readOnly: true
+{%- if pvc_name = "" %}
```
schmitch: maybe i need to make the check different?  is not null?

schmitch:
this should now ready for review and it fixes problems under gke, when creating a kadalu cluster since the google cloud containeros is readonly

aravindavk (approved):
Thanks for the PR. LGTM

Issue #303, opened 2020-09-21
Volumes are created but not mounted


robojones:
When I create multiple volume claims in a short timeframe, they are created but not mounted in the containers. I am pretty sure that this is an issue with kadalu. Any help is welcome!

Kadalu is configured as default storage class in my cluster. With small deployments using few volumes this works fine.
Now I am trying to install the [Harbor helmchart](https://github.com/goharbor/harbor-helm) which creates more than 10 pods with volume claims and every time some of them are stuck in the `ContainerCreating` phase. The volume claims are all created and bound, but attaching them to the containers fails.

To verify that this is not an issue with the Harbor helmchart, I created a simple deployment with 9 nginx containers and 9 PVCs.

**Minimal example:** [deployment.yaml](https://gist.github.com/robojones/df924fe209f106c1bc26b066db1fa9dc#file-deployment-yaml)

The pods which stuck in the `ContainerCreating` state always log the events:
- `FailedScheduling` with message `error while running "VolumeBinding" filter plugin for pod "...": pod has unbound immediate PersistentVolumeClaims`
- Later followed by:
`FailedMount` with the message `Unable to attach or mount volumes: unmounted volumes=[...]: timed out waiting for the condition`

**Full event log from kubectl describe:** [events.txt](https://gist.github.com/robojones/df924fe209f106c1bc26b066db1fa9dc#file-events-txt)

amarts:
Thanks for the report. Can you check the logs of `kadalu-csi-nodeplugin` pods, and see if you find any issues there?

robojones:
@amarts the kadalu-csi-nodeplugin pods have been running for the past five days. The logs look fine to me but I have uploaded them [here](https://gist.github.com/robojones/2a9b1d4c5bd541abe365ace98adb090d), so you can take a look.

amarts:
Interesting, the volume Attach is an operation to be handled by nodeplugin, so not all PVC requests came to the nodeplugin (or this particular nodeplugin. Can you check if there are kadalu-nodeplugin running in all nodes which needs the mount? there should be just 1 kadalu csi provisioner , but per node 1 nodeplugins. (if not, then there is an issue), and the mount issue may be from only few nodeplugins.

robojones:
I checked and `kadalu-nodeplugin` is running on all of my nodes. Here is a screenshot of my Rancher dashboard.
[...]

PR #304, opened 2020-09-22
across: use kubectl apply, instead of create

amarts:
If we create a pod (any instance) with kubectl 'create', we can't update it
with newer changes. It is good to use 'kubectl apply' inside the scripts.

Updates: #303


robojones:
I think this should also be updated in https://github.com/kadalu/kadalu/blob/devel/operator/main.py

robojones (approved):
lgtm

Issue #255, opened 2020-05-16
[RFE] Push Manifest files to Release instead of keeping in source tree


aravindavk:
Currently, manifest files are saved under `$SRC/manifests` directory and `kubectl kadalu install` uses files from the manifests directory.

- Add these manifest files to the release (Example: https://github.com/kadalu/kadalu/releases/tag/0.7.0)
- Change Documentation to use this path
- Change Kubectl kadalu CLI to use this path.

**Bonus**: Automatically copy the generated yaml files to release using Github actions.
https://github.com/marketplace/actions/upload-files-to-a-github-release

aravindavk:
Also use the same approach to upload the `kubectl-kadalu` binary.

Example: `KADALU_VERSION=0.7.1 make cli-build` and then copy `./cli/build/kubectl-kadalu` to release page.

amarts:
Done with `0.7.5` and now available in `0.7.6` too.

PR #313, opened 2020-09-28
Upload `kubectl-kadalu` binary to release page

aravindavk:
Updates: #255

### DIFF
--- a/.github/workflows/publish.yml
+++ b/.github/workflows/publish.yml
@@ -1,0 +1,41 @@
+name: Kubectl-kadalu and Manifest files
+
+on:
+  push:
+    # Publish `1.2.3` tags as releases.
+    tags:
+      - "*"
+
+jobs:
+  build:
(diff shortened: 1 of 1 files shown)


aravindavk:
Not sure how to test this

amarts (approved):
Should we add prepare-release also as part of this, which generates the operator files?

amarts:
> Not sure how to test this

We can make a alpha_release call it 0.7.2alpha

aravindavk:
> Should we add prepare-release also as part of this, which generates the operator files?

Ack. `KADALU_VERSION=${VERSION} make prepare-release-manifests` and copy all manifest files with built version.

aravindavk:
Currently copying all `yaml` files from `manifests` directory. Couldn't found a way to copy files of specific version. Looks like need to make changes to `Makefile` to generate new manifest files to separate directory.

amarts (approved):

PR #318, opened 2020-10-07
kubectl-kadalu: Support for k3s kubectl command

aravindavk:
k3s embeds the kubectl command and provides as subcommand. For
example to run `k3s kubectl get pods`.

All the kubectl-kadalu subcommands internally call `kubectl`
command to do the action. For example, `kubectl kadalu install`
will fetch the yaml file required and then calls `kubectl create -f <yaml>`

`--kubectl-cmd="k3s kubectl"` option now works properly while
using k3s. For example,

```
$ sudo k3s kubectl kadalu storage-list --kubectl-cmd="k3s kubectl"
```

### DIFF
--- a/cli/kubectl_kadalu/install.py
+++ b/cli/kubectl_kadalu/install.py
@@ -54,7 +54,7 @@ def run(args):
         operator_file = "%s/kadalu-operator%s%s.yaml" % (file_url, insttype, version)
 
     try:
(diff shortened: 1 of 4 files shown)


amarts:
Should we add corresponding Docs?

amarts (approved):

Issue #321, opened 2020-10-15
[Bug] sqlite3 Missing: kubectl kadalu storage-list --status


evanreichard:
When running, `kubectl kadalu storage-list --status`, the following is displayed:

> error 1 kubectl exec [POD] [COMMAND] is DEPRECATED and will be removed in a future version. Use kubectl exec [POD] -- [COMMAND] instead.
error: Internal error occurred: error executing command in container: failed to exec in container: failed to start exec "48a1357ecc82df899909b529df3add2d01be9e00dfec7dde1fa37e8834ecface": OCI runtime exec failed: exec failed: container_linux.go:370: starting container process caused: exec: "sqlite3": executable file not found in $PATH: unknown

It appears that the server Docker image no longer contains the sqlite3 binary needed to [execute the query](https://github.com/kadalu/kadalu/blob/d14fb2c0088b4bda3e341dec479cf8ed3dfac525/cli/kubectl_kadalu/storage_list.py#L196-L202)

aravindavk:
Thanks for reporting the issue. Looks like the base image is not rebuilt.

amarts:
@evanreichard #323 should have fixed the issue... along with #325 

Is the issue fixed for you?

amarts:
please use 0.7.2 version and above.

PR #323, opened 2020-10-19
Fix the kubectl exec deprecated message

aravindavk:
```
error 1 kubectl exec [POD] [COMMAND] is DEPRECATED and will be
removed in a future version. Use kubectl exec [POD] -- [COMMAND] instead.
```

Updates: #321


ook (approved):
For readability, I recommand to always put the namespace (-nkadalu) just after the `kubectl` command, but that's ok.

PR #331, opened 2020-11-07
Doc: provide information about yaml file based setup

amarts:
There were many questions about how to start with kadalu without python dependency on the host machine. Quite naturally, it was the default option, but we provided `kubectl kadalu ` subcommand thinking it would help many users. Now, adding the documentation to provide the information about yaml based approach to start kadalu operators.

### DIFF
--- a/doc/quick-start-yaml.md
+++ b/doc/quick-start-yaml.md
@@ -1,53 +1,26 @@
-# Quick Start
+# Quick Start with yaml files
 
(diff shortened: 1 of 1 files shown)


amarts:
Considering this is a doc improvement, which helps people comfortable with k8s ecosystem to use kadalu, will push it in. Lets improve based on feedback.

PR #334, opened 2020-11-09
Quotad: Fix for 64-bit inode XFS

Joibel:
kadala-quotad uses inode numbers as project numbers when setting up
quotas on XFS. Project numbers are always 32bit values, but when using
large disks XFS will generate 64 bit inode numbers.

This change modulos the inode number by 2^32 to get the inode number
in the correct range for project numbers. It does nothing good if this
results in a clash with another project number. As this seems to be
quite unlikely it feels good enough.

### DIFF
--- a/server/kadalu_quotad/quotad.py
+++ b/server/kadalu_quotad/quotad.py
@@ -27,13 +27,15 @@ except ImportError:
 #
 CONFIG_FILE = "/var/lib/glusterd/kadalu.info"
 
+PROJECT_MOD = 4294967296 # XFS project number is 32bit unsigned
 
 def set_quota(rootdir, subdir_path, quota_value):
     """
     Set Quota for given subdir path. Get project
-    ID from directory inode
+    ID from directory inode. XFS can have 64 bit inodes
+    so strip it to 32 bit and hope not to clash.
     """
-    ino = os.lstat(subdir_path).st_ino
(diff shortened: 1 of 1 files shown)


aravindavk:
Interesting find. One possible issue with this approach is that if two dirs with inode numbers `32330318` and `4327297614` will get the same project ID.

Joibel:
Yes, this is definitly a risk, although unlikely with the relatively small number of volumes per brick relative to the 2^32 possible projects. I haven't looked into how XFS allocates inode numbers, so perhaps that increases the risk.

Alternatives would rely on much more complex code detecting collisions and then mitigating them in a consistent and predictable way (remembering that the item you may have been colliding with in the past may have gone away), persisting data in some way, or generating project numbers in a completely diffferent way.

My preference would be to add a project number to the per-subvol .json file, and get the operator to allocate them, as it's the one with all the information and access to all the state. This is obviously a compatibility breaking and much more involved change.

amarts:
This fix is good :+1: 

In any case, ideal fix for us as project is to get glusterfs natively supporting quota, so we can use distribute. In which case, complete storage array can be a single glusterfs volume. Trying that with glusterfs quota.

This is required till that gets into glusterfs.

amarts (approved):

amarts:
@aravindavk feel free to merge when you have a look on this again.

aravindavk (approved):

PR #337, opened 2020-11-12
Automatic detection of bricks for quotad

Joibel:
I'm lazy and can't be bothered with setting up a json file or
environment variable for my bricks.

In the event that quotad cannot find anything to crawl via these
mechanisms, this change causes quotad to query gluster for all
volumes and all bricks on those volumes, to discover any that are on
this machine. If the brick name and hostname for some reason don't
match, this won't work. It would be normal that they would match.

Then we check for the 'kadalu' signature on those bricks of an info
and subvol directory in the root. If that is the case, crawl those
bricks.

The search for bricksprocess is time consuming, and so we only perform
it once every 30 crawls (about once a minute). This is not
configurable, just as the quota check period isn't.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -42,7 +42,7 @@ jobs:
   include:
     - name: pylint tests
       script:
-        - pip install requests xxhash grpcio pyxattr kubernetes
+        - pip install requests xxhash grpcio pyxattr kubernetes glustercli
         - make pylint || travis_terminate 1;
     - name: kadalu with kube 1.15.0
       script:
--- a/server/kadalu_quotad/glusterutils.py
+++ b/server/kadalu_quotad/glusterutils.py
@@ -1,0 +1,53 @@
+"""
+Utilities for reading information from gluster for 'external' quotad
+"""
+import os
+from glustercli.cli import volume
+
+KADALU_PATHS = {'info', 'subvol'}
+UUID_FILE = "/var/lib/glusterd/glusterd.info"
+
+myuuid = None
+
+
+def get_node_id():
+    """
+    Returns the local glusterd's UUID
+    """
+    global myuuid
+
+    if myuuid is not None:
+        return myuuid
+
+    val = None
+    with open(UUID_FILE) as uuid_file:
+        for line in uuid_file:
+            if line.startswith("UUID="):
+                val = line.strip().split("=")[-1]
+                break
+
+    myuuid = val
+    return val
+
+
+def get_automatic_bricks():
+    """
+    Returns array of paths to gluster bricks hosted on _this_ server
+    that appear to contain kadalu data and are therefore worth crawling
+    """
+    local_uuid = get_node_id()
+    found_bricks = []
+    for vol in volume.vollist():
+        for brick in volume.info(vol)[0]['bricks']:
+            if brick['uuid'] != local_uuid:
+                continue
+            brick_path = brick['name'].partition(':')[2]
+            is_kadalu_brick = True
+            for kadalu_path in KADALU_PATHS:
+                if not os.path.isdir(brick_path + '/' + kadalu_path):
+                    is_kadalu_brick = False
+                    break
+            if is_kadalu_brick:
+                found_bricks.append(brick_path)
+
+    return found_bricks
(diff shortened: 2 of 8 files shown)


amarts (changes requested):
Code looks good. Just a thought that it would be great to not have this when I deploy quotad inside kadalu pods.

2 things comes to mind:

1. Provide it under an option, if present, let this code path get executed.
2. Move relevant functions/methods to another file, so we are very clear this depends on external gluster dependency.

Asking this because in general, no other part of kadalu project depends on gluster cli (or glusterd) in general.

server/kadalu_quotad/quotad.py
```diff
@@ -26,6 +29,76 @@
...
+def get_bricklist(vol):
+    """
+    Returns array of bricks provided on _this_ gluster server
+    for a given volume
+    """
+    bricklist = glustercmd(["volume", "info", vol],
+                           'volInfo/volumes/volume/bricks/brick')
+    bricks = []
+    for brick in bricklist:
+        brick_match = re.match(HOSTNAME+':(.*)', brick)
```
aravindavk: Good to identify local bricks using node id, multiple IPs may fail in comparison for a few cases. `/var/lib/glusterd/glusterd.info` file will have this id. (Refer parsing here: https://github.com/gluster/glustercli-python/blob/1f26c606f1b107d0e92b5ba4ba66557330a2c98a/glustercli/metrics/utils.py#L25)

server/kadalu_quotad/quotad.py
```diff
@@ -130,19 +203,33 @@ def start():
...
+                try_automatic = False
                 config_data = json.loads(conf_file.read().strip())
                 for brick in config_data.get('bricks', None):
                     crawl(brick)
         except:    # noqa # pylint: disable=bare-except
             # Ignore all errors
             pass
 
+        if try_automatic:
+            # Getting the bricks is expensive, do it roughly
```
aravindavk: Volume info command is a local operation that reads from a local file in `/var/lib/glusterd`. The short interval should be fine.

aravindavk (commented):
Very nice enhancement. Thanks for your time.

server/kadalu_quotad/quotad.py
Joibel: Agree, that looks smarter.
I assume we don't want kadalu to depend on glustercli?

aravindavk: `glustercli` dependency is also fine if it works. I think it pulls in some extra dependencies like `paramiko` which is not useful here, but should be fine.

server/kadalu_quotad/quotad.py
Joibel: This process takes 3 seconds on a 'server class' machine which isn't particularly loaded but has 16 volumes with 5 bricks locally. I felt this was too much to do on every loop, so I've left it in.

server/kadalu_quotad/quotad.py
Joibel: Ok, great. Doing it that way saves on some code.

Joibel:
I've split out the new functions as requested.
It is now driven by BRICK_PATH=AUTO. I'm not sure this is needed as the current behaviour for an unconfigured quotad is to sit there doing no work, but appearing identical to one that is working.

server/kadalu_quotad/glusterutils.py
```diff
@@ -0,0 +1,53 @@
+"""
+Utilities for reading information from gluster for 'external' quotad
+"""
+import os
+from glustercli.cli import volume
```
amarts, line 5: mostly for tests to pass there may be a pip3 install requried in `.travis.yml`

amarts (approved):

aravindavk (approved):

.travis.yml
```diff
@@ -42,7 +42,7 @@ jobs:
   include:
     - name: pylint tests
       script:
-        - pip install requests xxhash grpcio pyxattr kubernetes
+        - pip install requests xxhash grpcio pyxattr kubernetes pythoncli
```
amarts: is it `glustercli` ?

Joibel: Doh. Yes. Travis didn't seem too upset by me typing rubbish though.

amarts (approved):

PR #338, opened 2020-11-12
Quotad json handling

Joibel:
Fix example kadalu.info file in the code and make quotad tell you when it cannot understand your kadalu.info file

### DIFF
--- a/server/kadalu_quotad/quotad.py
+++ b/server/kadalu_quotad/quotad.py
@@ -20,7 +20,7 @@ except ImportError:
 #    "bricks": [
 #      "/data/brick1",
 #      "/data/brick2"
-#    ],
+#    ]
 # }
 # We are using a json file, so the same file may get more parameters
 # for other tools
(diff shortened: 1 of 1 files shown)


Joibel:
@amarts Is the 1.15 test randomly failing. Build #620 (https://travis-ci.com/github/kadalu/kadalu/builds/200415474) passed for this. The only difference for this build is me remembering to signoff. I can't see where the 1.15 failure is in the sea of text.

Also k8s 1.15 is now deprecated, so we should be testing against 1.16 instead I think.

amarts (approved):

amarts:
> Also k8s 1.15 is now deprecated, so we should be testing against 1.16 instead I think.

Ack, I was thinking of moving to 1.17 and 1.20 once 1.20 is available.

About the current failure of 1.15, i saw that it is due to the PR #333... It is failing external glusterfs deployment as hostname was not reachable. While that is a good fix, seems like there are some flickering effect. I will keep that as a log information, and continue for now, instead of returning error.

Issue #330, opened 2020-11-02
replica3 recovery after 1 device unavailable


galp:
Hello,
I have been running happily kadalu on a small arm64 cluster and I like the project very much. 
 
However now I have an issue which I think should be documented a bit better

I have a replica3 storage pool on my cluster. One of the 3 nodes part of the pool is gone and the storage device on it will move to a new node.

First of all it seems that my  pods cannot work with two out of three replicas. That shouldnt be the case right?
My pods are stuck  creating with errors like this : 

`MountVolume.SetUp failed for volume "pvc-5bcb3600-d993-4e76-9b47-ca04ba70d6bd" : rpc error: code = Unknown desc = Exception calling application: [32]  mount: /var/lib/kubelet/pods/3cbe0ba9-059e-48cf-b09a-8a106805cc4a/volumes/kubernetes.io~csi/pvc-5bcb3600-d993-4e76-9b47-ca04ba70d6bd/mount: mount(2) system call failed: Transport endpoint is not connected.`
There is a possibility that the replicaset was already working with 2/3 replicas but I never noticed and now losing another killed it. 
Can you please explain a bit or perhaps point me to to the right place to understand how this works?

Also I would like to to know how to get the 3rd replica rebuild on a new node. What I am going to try in a minute when my new node is ready is to edit the relevant storage like 

` kubectl edit   KadaluStorage flash32 `
and change the node in there
Is this the way to recover?

Thank you very much.

amarts:
Good to hear that kadalu is working for you on arm64.

About replica3, it should technically work when only 2 nodes are available. Is there any hint in `csi-provisioner` &  `csi-nodeplugin` logs? (`-nkadalu`)

We will get back to you on the right way to edit the config (to make changes to volume).

galp:
Ok good. I think It might be beyond saving by now but it would be great to know how to recover in such cases.

amarts:
Ok, we now (version 0.7.5) have support for `handle_modified()` (ie, an option to change the storage config). Please test the behavior, and let us know if that is good enough.

Will close the issue for now, re-open if there are concerns.

amoravek:
I have the same problem on with kadalu 1.2.0 - it seems that every time i need to drain k8s node running the *first* replica, the pods using kadalu storage pvc remain in Terminating state forever. Some pods drained ok, but fail to initialize and log the following events:

` 
  Type     Reason       Age                   From               Message                                                                                                                                            
[...]

PR #343, opened 2020-11-16
operator: handle 'modified' operation

amarts:
With this, users can do a 'replace-brick' like operation for their replica3 or replica2 volume.

Notice that, for `Replica1` and `External` type, it wouldn't be supported, as it doesn't make sense.

Updates: #330


aravindavk (approved):
Looks good. Clients needs to reconnect or regenerate Volfile from config map to get the updated server lists.

amarts:
> Clients needs to reconnect or regenerate Volfile from config map to get the updated server lists.

'regenerate' is not required, as client volume file remains same (including the `remote-host` option, and device/brick details are for only server. And from the logs confirmed that k8s restarted the pod which got the device changed. Other pods continued to work without issues. 

So, should be good IMO.

PR #345, opened 2020-11-20
images: move the glustercli to server image (from base image)

### LINKED ISSUE #344: buildx actions failing for arm64 after making glustercli dependency in base image.
Few things we can try to fix:

* move this dependency only to kadalu-server Docker image.
* try to build it on an arm machine and see what is the error, and fix it in .github/workflow scripts.

amarts:
Server image already has major dependency w.r.to `glusterfs-server`
installed, and it makes sense that way.

Also make sure we have 'cli-build' done in scripts before building
images so `kubectl-kadalu` can be ready to be installed.

Also fix the other build failures due to `cli-build` dependency in operator image.

Fixes: #344

### DIFF
--- a/.github/workflows/dockerimage.yml
+++ b/.github/workflows/dockerimage.yml
@@ -22,6 +22,15 @@ jobs:
         with:
           buildx-version: latest
           qemu-version: latest
+      - name: Set up Python (for cli-build)
+        uses: actions/setup-python@v1
+        with:
+          python-version: '3.x'
+      - name: Get the CLI build (kubectl-kadalu)
+        run: |
+          python -m pip install --upgrade pip
+          pip install -r requirements.txt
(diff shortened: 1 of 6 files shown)


amarts:
@Joibel please have a look.

I had to make `glustercli` optional dependency as it failed install in arm64 builds.

Joibel:
Ok, makes sense.

I can remove the glustercli dependency and go back to scraping gluster --xml commands directly if we'd prefer that. I have all the code for it, it's just 'more code' that's mostly similar to what glustercli does better.

amarts:
> I can remove the glustercli dependency and go back to scraping gluster --xml commands directly if we'd prefer that. 

No, lets keep this for now. lesser code is surely preferred. We will wait till someone asks this on ARM64. Also we can push the pypi with arm64 support, so this issue wouldn't happen.

PR #354, opened 2020-12-07
csi: fix bug/typo in ext_volume index

bholmgren:
This patch fixed a typo that causes an exception when indexing ext_volume for external gluster volumes


amarts (approved):
Thanks.

aravindavk (approved):

Issue #314, opened 2020-09-29
Need for Generic improvements across codebase


amarts:
In a project, as the time passes some of the assumptions made at the beginning should be re-looked, analyzed, and kept up-to-date. Without which the project would grow and the tech backlog would remain.

This is an issue to capture all such efforts, so if we have any PRs linked here, you can assume that is a clean up patch :+1:

PR #363, opened 2020-12-14
fix buildx tests

amarts:
noticed that with buildx there were tests failing on github actions..

With these fixes, the tests are running fine!

Updates: #314


amarts:
@aravindavk this PR needs to be checked properly for k8s version bump up.. mainly as it caused some issues when we upgraded to 12.0.0 previously.

PR #365, opened 2020-12-15
operator: Remove storage_list dependency from kadalu-operator

### LINKED ISSUE #366: RFE: Remove storage_list dependency from kadalu operator
Current implementation of **handle_deleted()** makes use of **Storage** and **StorageUnit** classes of **storage_list.py**. 
This makes use of avoidable traversal of all available storages in **Storage** and **StorageUnit**. 

The implementation can be changed to get **_volname_** from **_obj_** and rest all the related information directly from __kadalu configmap__ saving initialization and retrieval traversal time.

vatsa287:
This PR removes storage_list dependency from the kadalu-operator.

Allowing to avoid Storage and StorageUnit intialization.

Gets the storage to be deleted from 'obj' and rest of the related information
directly from kadalu configmap. Reducing the current traversal time on Storage and StrorageUnit.

Fixes: #366 

### DIFF
--- a/operator/main.py
+++ b/operator/main.py
@@ -14,7 +14,6 @@ from jinja2 import Template
 from kubernetes import client, config, watch
 
 from kadalulib import execute as lib_execute, logging_setup, logf, send_analytics_tracker
-from storage_list import list_storages
 from utils import execute as utils_execute, CommandError
 
 NAMESPACE = os.environ.get("KADALU_NAMESPACE", "kadalu")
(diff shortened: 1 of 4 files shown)


operator/main.py
```diff
@@ -563,50 +559,31 @@ def get_num_pvs(storage):
...
 
     except CommandError as msg:
         logging.error(logf(
             "Failed to get size details of the "
-            "storage \"%s\"" % storage.storage_name,
+            "storage \"%s\"" % volname,
             error=msg
         ))
+        # Set default as 0
+        return 0
```
vatsa287, line 573: When PVC is not provisioned from that volume and is deleted, this block is executed but deletion stops. Hence setting default value as 0.

amarts:
It is a good practice to have an issue as reference for any PR... It may be an RFE, or bug.

vatsa287:
> It is a good practice to have an issue as reference for any PR... It may be an RFE, or bug.

Ill open it now.

Issue #364, opened 2020-12-14
Failure to create PVC


zzzhangqi:
System environment：centos arm64，k8s-1.18
**First question:**
```shell
#  I create it this way
kubectl kadalu storage-add rainbond-storage-pool-4 \
--device master-01:/dev/vdb
```
master-01 is an existing node

pods:
```shell
$ kubectl get pod -n kadalu
NAME                                 READY   STATUS    RESTARTS   AGE
kadalu-csi-nodeplugin-ll78d          3/3     Running   0          43m
kadalu-csi-nodeplugin-p8bwl          3/3     Running   0          43m
kadalu-csi-nodeplugin-rv42r          3/3     Running   0          43m
kadalu-csi-nodeplugin-rwdjk          3/3     Running   0          43m
kadalu-csi-nodeplugin-vhc4f          3/3     Running   0          43m
kadalu-csi-nodeplugin-xcd7p          3/3     Running   0          43m
kadalu-csi-provisioner-0             4/4     Running   0          43m
operator-76b85b86b6-dtwds            1/1     Running   0          43m
server-rainbond-storage-pool-4-0-0   2/2     Running   0          11m
```
```shell
#error logs :

Events:
  Type     Reason                Age                   From                                                                  Message
  ----     ------                ----                  ----                                                                  -------
  Warning  ProvisioningFailed    112s                  kadalu_kadalu-csi-provisioner-0_2e0f6b97-789b-488d-a866-aa6682151aa4  failed to provision volume with StorageClass "kadalu.replica1": rpc error: code = DeadlineExceeded desc = context deadline exceeded
  Normal   Provisioning          111s (x2 over 4m52s)  kadalu_kadalu-csi-provisioner-0_2e0f6b97-789b-488d-a866-aa6682151aa4  External provisioner is provisioning volume for claim "default/webapp-pv"
  Normal   ExternalProvisioning  3s (x21 over 4m52s)   persistentvolume-controller                                           waiting for a volume to be created, either by external provisioner "kadalu" or manually created by system administrator
```

**Second question:**
```shell
kubectl kadalu storage-add rainbond-storage-pool-5 --path master-01:/root/data/

#error logs:
Events:
  Type     Reason       Age               From               Message
  ----     ------       ----              ----               -------
  Normal   Scheduled    33s               default-scheduler  Successfully assigned kadalu/server-rainbond-storage-pool-5-0-0 to rainbond-master-01
[...]

PR #371, opened 2020-12-18
RKE: fix the KUBELET_DIR path

amarts:
Updates: #364


amarts:
Also from Slack:

> @ Sebastien Deprez  5:36 PM
> there's one error in the manifests for RKE (v2.5.1) and that's the path for KUBELET_DIR which is /var/lib/kubelet instead of the one in the current files.

aravindavk (commented):
Looks good. But these changes will be overwritten by https://github.com/kadalu/kadalu/blob/devel/extras/scripts/gen_manifest.py#L37

Issue #55, opened 2019-11-04
RFE: provide helm chart


amarts:
[helm charts](https://helm.sh/) are a good way to have more reach in k8s world. Would be good to have a helm chart implemented for kadalu project.

amarts:
More on it here : https://docs.bitnami.com/kubernetes/how-to/create-your-first-helm-chart/

amarts:
we don't have helm chart yet, and we still need one!

unixfox:
Bump, still interested.

amarts:
basic helm chart is provided in the repo now. Will keep tracking the issues separately for any specific bugs from now on.

PR #375, opened 2020-12-18
helm chart support

amarts:
* Create a helm chart dir
* Add all things from kadalu-operator.yaml to helm chart

How it can be worked out:

```
cd helm
kubectl create namespace kadalu # <- currently this is required
helm install --namespace kadalu kadalu-chart kadalu/ --values kadalu/values.yaml
```

With this PR, we can keep our helm chart open for further contributions.

Updates: #55

### DIFF
--- a/helm/kadalu/.helmignore
+++ b/helm/kadalu/.helmignore
@@ -1,0 +1,23 @@
+# Patterns to ignore when building packages.
+# This supports shell glob matching, relative path matching, and
+# negation (prefixed with !). Only one pattern per line.
+.DS_Store
+# Common VCS dirs
+.git/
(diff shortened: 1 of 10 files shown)


amarts:
Haven't posted in any helm repositories yet, happy to hear which is the common place some of our users use right now.

aravindavk:
👏 

We can also add `--dist <k8s-dist>` and support customizations(As we are doing now with multiple manifest files)

amarts:
> We can also add --dist <k8s-dist> and support customizations(As we are doing now with multiple manifest files)

Makes sense. I will start opening issues and fixing one after the other.

Issue #393, opened 2020-12-23
Python exception "TypeError: unhashable type: 'list'" with external GlusterFS on 0.7.4


floek:
Hi,

after fixing #376 I now get another Python exception when executing kubectl kadalu storage-list:
```
# kubectl kadalu storage-list
Traceback (most recent call last):
  File "/usr/lib/python3.8/runpy.py", line 194, in _run_module_as_main
    return _run_code(code, main_globals, None,
  File "/usr/lib/python3.8/runpy.py", line 87, in _run_code
    exec(code, run_globals)
  File "/usr/local/bin/kubectl-kadalu/__main__.py", line 64, in <module>
  File "/usr/local/bin/kubectl-kadalu/__main__.py", line 53, in main
  File "/usr/local/bin/kubectl-kadalu/storage_list.py", line 235, in run
  File "/usr/local/bin/kubectl-kadalu/storage_list.py", line 72, in list_storages
TypeError: unhashable type: 'list'
```

Cheers,

floek

amarts:
@vatsa287 fyi... see if you can add it as a test in tests/minikube.sh That way, we can make sure the changes are immediately confirmed to be working!

vatsa287:
sure I'll add it

floek:
This still exists in 0.7.5:
```
# kubectl-kadalu version
kubectl-kadalu 0.7.5
# kubectl kadalu storage-list
Traceback (most recent call last):
  File "/usr/lib/python3.8/runpy.py", line 194, in _run_module_as_main
    return _run_code(code, main_globals, None,
  File "/usr/lib/python3.8/runpy.py", line 87, in _run_code
    exec(code, run_globals)
  File "/usr/local/bin/kubectl-kadalu/__main__.py", line 64, in <module>
  File "/usr/local/bin/kubectl-kadalu/__main__.py", line 53, in main
  File "/usr/local/bin/kubectl-kadalu/storage_list.py", line 235, in run
  File "/usr/local/bin/kubectl-kadalu/storage_list.py", line 72, in list_storages
TypeError: unhashable type: 'list'
```

PR #397, opened 2020-12-23
temporarily disable arm builds

amarts:
* add a separate automated build for testing arm builds
* This is expected to last for another minor release

Updates: #393 


amarts:
@aravindavk Planning to fix arm builds separately for a week... For now, arm latest release support will be 0.7.2 and will enable 'arm' tag as a separately available... Will update the README with this.

Issue #328, opened 2020-11-01
Support for armv7 


gabrielruiu:
The kadalu docker images currently have support for arm64, which is great, but it would be cool to also add support for armv7. I am currently trying to deploy a kubernetes cluster where some of the nodes are odroids (HC1 and HC2) which are running on a armv7 processor. Right now it's a blocker in my case.

amarts:
@gabrielruiu is it possible to build on your architecture with just docker build with kadalu source code. I am facing some issues with arm/v7 tag (#329)... if it works locally, then we can consider automating later!.

sputnik13:
@amarts I was able to do `make release` natively on an arm32v7 (odroid-hc1 + armbian buster), however there were missing dependencies in the Dockerfile (addressed in #402)

also there's a known issue with using ubuntu:20.04 on arm32v7 (https://github.com/moby/moby/issues/40734) which may impact ability to build depending on whether the build agent is hosted on ubuntu/focal + docker.io or debian/buster + docker-ce

debian/buster + docker-ce can successfully build if libseccomp2 is updated via buster-backports, but ubuntu/focal currently does not seem to work ootb

amarts:
`0.7.6` and beyond will have support for `arm/v7`. Please use.

Thanks to @sputnik13

PR #402, opened 2020-12-24
add dev dependencies to Dockerfiles

sputnik13:
adding libffi-dev, build-essential, python3-dev, and libssl-dev to
[csi,operator,server]/Dockerfile to allow successful build on arm32v7

Updates: #328


amarts:
Thanks for the PR, can you also check if you can edit https://github.com/kadalu/kadalu/blob/devel/.github/workflows/on-pr-merge_arm.yml to include the type linux/armv7h in script ? so it can be automatically built from next time.

amarts:
Tried it at -- https://github.com/kadalu/kadalu/commit/06e64fefc1c59b9ae71694a410920b0ae20cd4fa

Issue #377, opened 2020-12-19
Upgrade guide / documentation


floek:
Hi,

how do I upgrade my kadalu pods to the new 0.7.3 version? I found no documentation.
In the meantime, I reinstalled.

Thanks,
Floek

floek:
I still wonder how to upgrade properly (#399). I'm at a testing environment now and do:
- delete all Kadalu pvs
- delete all ressources (including kadalustorage objects) in the kadalu namespace
- delete all Kadalu storageclasses
- delete the kadalu executable
- delete all data on the external Gluster volume
- run my ansible playbook and redeploy kadalu

But I think this is not the way to do it in production ;-).

floek:
Hi, how do I upgrade to 0.7.7?

ook:
Hi @floek . Not an authoritative answer but here "my" way to upgrade:

* run script https://github.com/kadalu/kadalu/blob/devel/extras/scripts/cleanup
* follow the installation instruction from https://github.com/kadalu/kadalu/blob/devel/doc/quick-start.md
* pray^W fix the issues by hand, case by case
HTH

Issue #352, opened 2020-11-30
upgrade guide


schmitch:
since kadalu is a storage platform it should have a upgrade guide, especially to not loose data.
even if it is just `kubectl kadalu upgrade` or upgrading deployment versions.
I mean what nobody wants is loosing data ;-)

Especially with changes that might be disruptive like: #305 i.e. an existing cluster might not upgrade the same the newer clusters.

amarts:
#242 #213 and #106 were opened with similar points.

amarts:
@schmitch adding a comment here for your reference. There is still more work pending to handle replica2/replica3 type of volumes.

With the latest version `0.7.6`, to upgrade all storage pods, along with operator, we just have to pass the latest operator yaml file from one of latest releases to `kubectl apply -f`. That should automatically pick the newer version for CSI, Storage pods, and also the operator pods.

schmitch:
I will try it out in the next few weeks. Well we only use replica3 volumes, does this also work? or what exactly does mean "more work pending"?

amarts:
ok, for replica3 volume we need to do a 'rolling' upgrade. That is the work pending. Ie, the operator should pick the first storage node, upgrade it, and keep watching on heal pending, and once its 0, upgrade second, and watch for heal pending, and once its 0, proceed with node3. 

If there is no I/O during upgrade, then the current method of upgrading without checking for heal count would work without any issues.

attrib:
Tried to upgrade from 0.8.3 to 0.8.4, but `kubectl kadalu install` said `Kadalu operator already installed`

I believe in https://github.com/kadalu/kadalu/blame/devel/cli/kubectl_kadalu/install.py#L55 is a check missing, which version is installed, otherwise it will never get to the `kubectl apply -f` mentioned in https://github.com/kadalu/kadalu/issues/352#issuecomment-762647907

PR #411, opened 2021-01-09
Support upgrade of pods

amarts:
When the operator starts, check if config map is already present (in which case, its most probable case of upgrade). Just make sure to trigger `kubectl apply` on storage pods as they most probably need upgrade too.

Updates: #377 #352 and similar few

### DIFF
--- a/operator/main.py
+++ b/operator/main.py
@@ -32,7 +32,8 @@ VOLUME_TYPE_REPLICA_2 = "Replica2"
 VOLUME_TYPE_REPLICA_3 = "Replica3"
 VOLUME_TYPE_EXTERNAL = "External"
 
-CREATE_CMD = "apply"
+CREATE_CMD = "create"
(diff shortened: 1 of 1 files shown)


amarts:
Tests pending, and hence in draft mode. Will be happy if someone tries this upgrade process locally too. (git clone the branch, build one version, now use the new version built to replace the running (test infra) instance, and you should see even the storage pods to be updated).

amarts:
Did basic testing, and the upgrade is happening.

There are other issues which came up when I tested. The storage pods were not getting Terminated completely, and stayed in 'Terminating' state for more than 60secs (ie, the graceperiod), so I had to kill them manually (with `kubectl delete pod -nkadalu storage-pool-pod-name`), and the new pod with newer version came up.

The code to upgrade storage pods and CSI pods to latest version seems to be working fine :-) 

Will merge this and continue on issues separately!

PR #412, opened 2021-01-14
support adding multiple hosts as option in external storage config

amarts:
There is a need for option similar to gluster's `backup-volfile-servers` option in kadalu when an external storage is used. It is critical to provide high availability, where admin can peacefully give config once and forget about issues of availability of one of the nodes for mounts to work.

### DIFF
--- a/operator/main.py
+++ b/operator/main.py
@@ -70,20 +70,23 @@ def bricks_validation(bricks):
     return ret
 
 
-def is_host_reachable(host, port):
+def is_host_reachable(hosts, port):
     """Check if glusterd is reachable in the given node"""
     timeout = 5
     sock = socket.socket(socket.AF_INET, socket.SOCK_STREAM)
     sock.settimeout(timeout)
-    try:
-        sock.connect((host, int(port)))
-        sock.shutdown(socket.SHUT_RDWR)
-        return True
-    except socket.error as msg:
-        logging.error(logf("Failed to open socket connection", error=msg))
-        return False
-    finally:
-        sock.close()
+    for host in hosts:
+        try:
+            sock.connect((host, int(port)))
+            sock.shutdown(socket.SHUT_RDWR)
+            return True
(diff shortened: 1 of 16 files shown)


aravindavk (approved):
Looks good. How about supporting both so that currently deployed ones will work after upgrade.

```
hosts = data.get("gluster_hosts", [])
host = data.get("gluster_host", None)
if host is not None:
    hosts.append(host)

"g_hosts": hosts
```

amarts:
> kadalulib.CommandException: [1]  error: error validating "/kadalu/templates/external-storageclass.yaml": error validating data: ValidationError(StorageClass.parameters.gluster_hosts): invalid type for io.k8s.api.storage.v1.StorageClass.parameters: got "array", expected "string"; if you choose to ignore these errors, turn validation off with --validate=false

aravindavk:
> kadalulib.CommandException: [1] error: error validating "/kadalu/templates/external-storageclass.yaml": error validating data: ValidationError(StorageClass.parameters.gluster_hosts): invalid type for io.k8s.api.storage.v1.StorageClass.parameters: got "array", expected "string"; if you choose to ignore these errors, turn validation off with --validate=false

Oh then we need to use string with `,` separated.

operator/main.py
```diff
@@ -391,16 +398,31 @@ def handle_external_storage_addition(core_v1_client, obj):
...
         "volume_id": obj["spec"]["volume_id"],
         "type": VOLUME_TYPE_EXTERNAL,
         "kadalu-format": True,
-        "gluster_host": details["gluster_host"],
+        "gluster_hosts": json.dumps(hosts),
         "gluster_volname": details["gluster_volname"],
         "gluster_options": details.get("gluster_options", "ignore-me"),
     }
 
+    logging.info(logf("DEBUG - data", data=data))
```
aravindavk: `logging.debug`

PR #413, opened 2021-01-18
change CRD version from v1beta1 to v1

### LINKED ISSUE #388: Warning seen while running operator
```
Warning: apiextensions.k8s.io/v1beta1 CustomResourceDefinition is deprecated in v1.16+, unavailable in v1.22+; use apiextensions.k8s.io/v1 CustomResourceDefinition
```

amarts:
Fixes: #388

### DIFF
--- a/cli/kubectl_kadalu/storage_add.py
+++ b/cli/kubectl_kadalu/storage_add.py
@@ -173,12 +173,10 @@ def storage_add_data(args):
     # External details are specified, no 'storage' section required
     if args.external:
         node, vol = args.external.split(":")
-        content["spec"]["details"] = [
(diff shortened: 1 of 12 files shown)


amarts:
Note that this change is not backward compatible with external storage config file. But I feel the change is worth doing because an array for 'details' doesn't makes sense, and we now have proper schema published for users to cross check.

Plan is to make a new release soon, so all new deployments can use new format.

amarts:
Note to those who have things installed. This change doesn't harm the 'deployed' setup, as this is just the change in the 'input' file format. Comes to effect only during adding new storage config.

PR #421, opened 2021-01-20
cli: use the release version as default while using release package

### LINKED ISSUE #418: Kubectl-kadalu release should deploy versioned operator
After downloading kubectl-kadalu and running `kubectl kadalu install` the kadalu-operator.yaml that is used is: https://raw.githubusercontent.com/kadalu/kadalu/devel/manifests/kadalu-operator.yaml which points to a devel image. A release versioned kubectl-kadalu should install that version of the kadalu operator. For example the 0.7.6 version should install using the release asset: https://github.com/kadalu/kadalu/releases/download/0.7.6/kadalu-operator.yaml.

This effects arm deployments as the devel image isn't multiarch compiled while the release versions are.

amarts:
Fixes: #418 

### DIFF
--- a/cli/kubectl_kadalu/install.py
+++ b/cli/kubectl_kadalu/install.py
@@ -16,9 +16,9 @@ def set_args(name, subparsers):
 
     arg(
         "--version",
-        help="Kadalu Version to Install [default: latest]",
-        choices=[VERSION, "devel", "latest"],
-        default="latest"
+        help="Kadalu Version to Install [default: " + VERSION + "]",
+        choices=[VERSION],
+        default=VERSION
     )
     arg(
         "--type",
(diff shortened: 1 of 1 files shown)


cli/kubectl_kadalu/install.py
```diff
@@ -63,11 +63,11 @@ def run(args):
 
     operator_file = args.local_yaml
     if not operator_file:
-        file_url = "https://raw.githubusercontent.com/kadalu/kadalu/devel/manifests"
+        file_url = "https://github.com/kadalu/kadalu/releases/download/%s" % VERSION
```
aravindavk: good to keep `args.version` instead of `VERSION` here.

PR #426, opened 2021-01-29
Dockerfile: follow the advised method to reduce the image size

amarts:
Build the base image:

`sudo docker build -t kadalu/builder:latest -f extras/Dockerfile.builder .`

Few things which changes with this commit:

* use kadalu-storage (glusterfs fork of kadalu, which adds features like simple-quota etc).
* changes in templates to accomodate distribute volumes
* remove usage of `-oprjquota` in mount
* use simple quota limit feature to set the quota
* simplify the brick volfile generation, and make it uniform across
* provide 'client-pid' to clients mounted in csi controller (for simple-quota)
* fix the template files
* changes to include quota-crawler with provisioner (Dockerfile & script included)

So what are the benefits for users:

* Instead of creating multiple host volumes of same type, they can use just one volume with all bricks
  - Scalling of the host volume is just adding storage to config and calling `kubectl apply -f` again.
* Any backend filesystem which supports extended attributes can be used on devices.
* no requirement of 'quotad' as an container in server pods
* Image sizes of all gluster pods would reduce further!


aravindavk:
Changes looks great. Will review once again by tomorrow. 👍

extras/quota-crawler.sh
```diff
@@ -0,0 +1,24 @@
...
+    if [ ! -d $MOUNT_DIR/*/subdir ] ; then
+	echo "No PVC yet, continuing to watch..."
+	sleep 30;
+	continue;
+    fi
+
+    # Subdir is in the form /mnt/$host-volname/subdir/NN/MM/PVCNAME
+    for dir in $(find $MOUNT_DIR/*/subvol/*/* -maxdepth 1 -mindepth 1 -type d); do
+	used_size=$(df -B1 ${dir}  | tail -n1 | awk '{print $3}');
+	echo "$dir : $used_size"
```
amarts, line 17: can remove this after debugging

server/shd.py
```diff
@@ -19,6 +19,18 @@ def generate_shd_volfile(client_volfile, volname, voltype):
     with open(info_file_path) as info_file:
         data = json.load(info_file)
 
+    # Tricky to get this right, but this solves all the elements of distribute in code :-)
+    data['dht_subvol'] = []
+    if data["type"] == "Replica1":
```
amarts, line 26: this should always be `False` in shd.py.

aravindavk:
To support upgrade, We need to provide a script to run in csi-provisioner container. Crawl each host volumes, read info file and run simple quota xattrs commands.

extras/quota-crawler.sh
aravindavk: can be made conditional if `LOG_LEVEL`(or similar env var) is available.

```
if [[ $LOG_LEVEL == "DEBUG" ]]; then echo "Quota Total usage xattr $dir : $used_size"; fi
```

templates/csi.yaml.j2
```diff
@@ -405,13 +405,20 @@ spec:
...
           command: ["/bin/sh"]
           args: ["-c", "while true; do logcnt=$(/bin/ls /var/log/gluster/ | wc -l); if [ ${logcnt} -gt 0 ]; then break; fi; sleep 5; done; tail -f /var/log/gluster/*.log"]
           volumeMounts:
             - name: varlog
               mountPath: "/var/log/gluster"
+        - name: kadalu-quota-crawler
+          image: docker.io/kadalu/quota-crawler:latest
+          volumeMounts:
+            - name: mountpath
+              mountPath: /mnt
```
aravindavk: is this mount path available? Server template uses `Bidirectional` to make the subdirs visible.

https://github.com/kadalu/kadalu/blob/devel/templates/server.yaml.j2#L99

aravindavk:
Just a thought. How about running crawler and csi provisioner in single container?(Also `glusterfsd` and `shd` in single container itself). Main process starts multiple subprocess using `subprocess`. Advantage is no shared mounts required between container. This may resolve the terminate issue of Server pods due to `Bidirectional` brick mount.

Server container:

- Prepare the mount and required dirs
- Generate Brick and Shd Volfile.
- Start brick and shd as subprocess.
- Optionally, main process can monitor the sub processes.

CSI Provisioner:

- Start CSI provisioner and Crawler as subprocess.

amarts:
Makes sense! Good to experiment for sure!

aravindavk:
> Makes sense! Good to experiment for sure!

Realised that, it may become difficult to rate control(cpu, memory and other resources) the specific process if multiple processes exists in same container. Let me look for solving `Bidirectional` mount propagation issue.

amarts:
> Realised that, it may become difficult to rate control(cpu, memory and other resources) the specific process if multiple processes exists in same container. 

For the benefit it gets, and as all these CPUs/memory are part of backend storage process, it is OK IMO. Lets try merging these processes with subprocess and then see how we can change it later!

aravindavk:
> > Realised that, it may become difficult to rate control(cpu, memory and other resources) the specific process if multiple processes exists in same container.
> 
> For the benefit it gets, and as all these CPUs/memory are part of backend storage process, it is OK IMO. Lets try merging these processes with subprocess and then see how we can change it later!

Implemented here https://github.com/kadalu/kadalu/pull/432

extras/Dockerfile.simple-quota-crawler
```diff
@@ -0,0 +1,8 @@
+FROM alpine
+
+RUN apk add --no-cache bash
+RUN apk add --no-cache attr
+
+COPY quota-crawler.sh /quota-crawler.sh
```
aravindavk, line 6: Now this can be moved to csi Dockerfile and add the following in `csi/start.py`

```python
mon.add_process(Proc("quota-crawler", "/bin/bash", ["/quota-crawler.sh"]))
```

templates/csi.yaml.j2
```diff
@@ -405,13 +405,20 @@ spec:
...
         - name: kadalu-logging
           image: busybox
           command: ["/bin/sh"]
           args: ["-c", "while true; do logcnt=$(/bin/ls /var/log/gluster/ | wc -l); if [ ${logcnt} -gt 0 ]; then break; fi; sleep 5; done; tail -f /var/log/gluster/*.log"]
           volumeMounts:
             - name: varlog
               mountPath: "/var/log/gluster"
+        - name: kadalu-quota-crawler
```
aravindavk: This can be removed if we move quota crawler to csi container itself.(Ref comment above)

PR #430, opened 2021-02-04
Install latest version of GlusterFS

amarts:
Go with glusterfs-8 which is supported right now, and in later versions, change it to glusterfs-9.


sputnik13:
lgtm

PR #432, opened 2021-02-07
Process monitor and Prometheus exporter

aravindavk:


aravindavk:
Rebase merge is preferred.

amarts (approved):
This is overall a nice improvement. Helps us a lot. Also noticed that storage pod termination (and restart) are now smooth.

Went through the test logs, and see that all are good. Only thing which changed is, operator took some extra time to get started in this PR. Not a blocker for now, but will check it out for few more runs later.

PR #433, opened 2021-02-08
core: use kadalu-storage branch

amarts:
Build the base image:

`sudo docker build -t kadalu/builder:latest -f extras/Dockerfile.builder .`

Few things which changes with this commit:

* use kadalu-storage (glusterfs fork of kadalu, which adds features like simple-quota etc).
* changes in templates to accomodate distribute volumes
* remove usage of -oprjquota on mount
* use simple quota limit feature to set the quota
* simplify the brick volfile generation, and make it uniform across
* provide 'client-pid' to clients mounted in csi controller (for simple-quota)
* fix the template files
* include quota-crawler with provisioner

So what are the benefits for users:

* Instead of creating multiple host volumes of same type, they can use just one volume with all bricks
  - Scalling of the host volume is just adding storage to config and calling `kubectl apply -f` again.
* Any backend filesystem which supports extended attributes can be used on devices.
* no requirement of 'quotad' process
* Image sizes of all gluster pods would reduce further, as we use builder mechanism.


amarts:
all review comments and testing is done with #426

csi/start.py
```diff
@@ -10,6 +10,7 @@ def main():
     mon = Monitor()
     mon.add_process(Proc("csi", "python3", [curr_dir + "/main.py"]))
     mon.add_process(Proc("metrics", "python3", [curr_dir + "/exporter.py"]))
+    mon.add_process(Proc("quota", "bash", [curr_dir + "/quota-crawler.sh"]))
```
aravindavk: Same container image is pulled by `csi-provisioner` and `csi-nodeplugin`. To avoid starting this process in `nodeplugin`, we can add one env variable in csi.yaml.j2 template(Ex: `CSI_ROLE=provisioner`).

```
if os.environ.get("CSI_ROLE", "-") == "provisioner":
    mon.add_process(Proc("quota", "bash", [curr_dir + "/quota-crawler.sh"]))
```

aravindavk (commented):
LGTM.

This line can be removed now https://github.com/kadalu/kadalu/blob/devel/server/server.py#L31

aravindavk (approved):

aravindavk:
For upgrade support a separate script can be provided. That can be run inside csi provisioner node when needed.

For each `/mnt/*/subvol`

```
info_file = <hostvol_mnt>/info/<volpath>.json
with open(info_file) as inf:
    data = json.load(inf)
    try:
        retry_errors(os.setxattr,
                     [os.path.join(<hostvol_mnt>, <volpath>),
                      "trusted.glusterfs.namespace",
                      "true".encode()],
                     [ENOTCONN])
        retry_errors(os.setxattr,
                     [os.path.join(<hostvol_mnt>, <volpath>),
                      "trusted.gfs.squota.limit",
                      str(data["size"]).encode()],
                     [ENOTCONN])
    # noqa # pylint: disable=broad-except
    except Exception as err:
        logging.info(logf(
            "Failed to set quota using simple-quota. Continuing",
            error=err
            ))

```

This can be automated by triggering this to run whenever a Host volume is mounted(That is after this line https://github.com/kadalu/kadalu/blob/devel/csi/volumeutils.py#L214)

amarts:
@aravindavk will merge this and try to make a 'pre-release' from this, so we can experiment if everything is OK. Will revert if anything major comes up!

aravindavk:
> @aravindavk will merge this and try to make a 'pre-release' from this, so we can experiment if everything is OK. Will revert if anything major comes up!

Ack. LGTM

PR #445, opened 2021-02-13
some cli fixes for logs and healcheck

amarts:


amarts:
recommend 'rebase and merge' as each of the commit solves different problem.

PR #449, opened 2021-02-18
client volfile: add an option to reduce logs

amarts:
Updates: #314

### DIFF
--- a/templates/Replica2.client.vol.j2
+++ b/templates/Replica2.client.vol.j2
@@ -1,5 +1,6 @@
 volume {{ volume_id }}-ta
     type protocol/client
+    option transport.socket.read-fail-log false
     option remote-host {{ tiebreaker["node"] }}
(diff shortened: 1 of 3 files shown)


amarts:
This option makes logs like below go away:

```
[2021-02-16 11:25:15.830574 +0000] W [socket.c:764:__socket_rwv] 0-tcp.replica3-pool-server: readv on 10.42.1.1:50836 failed (No data available)
[2021-02-16 11:25:19.227333 +0000] W [socket.c:764:__socket_rwv] 0-tcp.replica3-pool-server: readv on 10.42.3.7:49148 failed (No data available)
```

Notice that here, while it feels like readv() failed on a remote connection, it is printing the peerinfo of local connection endpoint, hence different ports are seen in every such logs. This logs appear when a server is down. 

By adding this PR, such logs will not be seen, and would cause less confusion.

PR #454, opened 2021-02-22
Fixed error 8 in line 592 in /operator/main.py

manav-7:
Added line 615 to resolve the error.

### DIFF
--- a/operator/main.py
+++ b/operator/main.py
@@ -612,6 +612,7 @@ def get_configmap_data(volname):
             "Failed to get details from configmap",
             error=err
         ))
+        return None
 
 
 def get_num_pvs(storage_info_data):


amarts:
Welcome to the kadalu project @manav-7 :fist_right: 

Few suggestions:

* Always use `-s` option while doing a commit (so that it adds a Signed-off message in commit).
* Commit subject/title should be about what you have done, in which component. In this case it can be something like: "operator: fix the pylint error in main.py"
* Commit message body should contain details about the fix, what you have done etc: "with new version of pylint there was an error for not having 'return' as last statement, now fixed" or something similar.

This is a good practice in any project you want to contribute.

Just do a push with those changes to the same commit with `--amend` and then do `--force` push.

manav-7:
Thank you for the suggestions sir.
I will put them into practice from next time.👍

leelavg:
@manav-7 

One more suggestion in addition to above, as this PR is addressing an issue, in your commit message either add one of the below two:
`Fixes: <github_issue_url>` (if PR fixes the issue) (or)
`Updates: <github_issue_url>` (if PR partially fixes the issue)

manav-7:
Yes ma'am.

PR #459, opened 2021-02-24
Test helm template on every PR submit

leelavg:
Gains:
- Forces committer to update helm template if only .j2 is updated
- Clearly show in CI log about the differences in template vs manifest
- Soft test every distro helm template against operator manifest

Loss:
- Not explicitly testing installation of actual helm chart
- For false positives wrt helm, skip marker has to be supplied

Minor Fixes:
- Updated manifests by running `make gen-manifest` locally
- Changed VERBOSE environment variable to accept `yes/no` values

Local tests:
- Updates in bash script is tested locally
- YAML lint is run on edited files

NO changes to release directives


manifests/kadalu-operator-microk8s.yaml
```diff
@@ -282,7 +283,11 @@ spec:
...
+            - name: KADALU_NAMESPACE
+              value: "kadalu"
             - name: KUBELET_DIR
               value: "/var/snap/microk8s/common/var/lib/kubelet"
             - name: K8S_DIST
-              value: "microk8s"
\ No newline at end of file
+              value: "microk8s"
+            - name: VERBOSE
+              value: "False"
```
aravindavk: @amarts just noticed that, `"False"` here will result in adding verbose logs because all env variables are passed as string.  https://github.com/kadalu/kadalu/blob/devel/lib/kadalulib.py#L139

We should change this to `VERBOSE = yes|no` and compare the same inside kadalulib.py

Sorry @leelavg, issue is not related to your PR. But noticed the issue here.

leelavg: @aravindavk np, while raising this PR i noticed two other minor issues and added the fix. Actually I wanted to bring this to your notice coz, True/False in helm template is being forced to behave as a string because of VERBOSE env value.

I can rebase once the value is changed and committed.

leelavg: Fixed.

amarts (approved):

PR #465, opened 2021-03-03
doc: Fix kubectl-kadalu version

aravindavk:

### DIFF
--- a/doc/quick-start.md
+++ b/doc/quick-start.md
@@ -6,7 +6,7 @@
 Download the latest release of Kadalu Kubectl plugin using,
 
 ```
-curl -LO https://github.com/kadalu/kadalu/releases/download/0.7.7/kubectl-kadalu
+curl -LO https://github.com/kadalu/kadalu/releases/download/0.8.0/kubectl-kadalu
 ```
 
 Make the kubectl binary executable.
--- a/doc/setup.md
+++ b/doc/setup.md
@@ -5,14 +5,13 @@
 - Kubernetes 1.13.0 + version
 - The host should support xfs (`mkfs.xfs`)
   - On some systems this might require installation of xfsprogs package
-- The `mount -t xfs` with `-oprjquota` should work
 
 ### Setup
 
 Download the latest release of Kadalu Kubectl plugin using,
 
 ```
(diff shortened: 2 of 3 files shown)


amarts:
This is one thing I wanted to syncup with you. Is there a way we push a static content to kadalu.github.io repo, say, `kadalu.io/repo/kadalu/latest`, which would be updated from the release script and we can have release version always valid in docs by giving curl syntax with this link in release variable.

aravindavk:
We can copy to `gh-pages` of this repo once released. If we copy `kubectl-kadalu` to root dir of `gh-pages` branch then it will be available as https://kadalu.io/kadalu/kubectl-kadalu

aravindavk:
> We can copy to `gh-pages` of this repo once released. If we copy `kubectl-kadalu` to root dir of `gh-pages` branch then it will be available as https://kadalu.io/kadalu/kubectl-kadalu

Like this https://github.com/kadalu/docs/blob/master/.github/workflows/on-pr-merge.yml#L28-L33

amarts:
sorry, earlier was using phone to type, so couldn't explain in detail. 

What I was thinking was something like below:

```
curl -LO https://github.com/kadalu/kadalu/releases/download/$(curl -L -s https://kadalu.io/releases/kadalu/latest.txt)/kubectl-kadalu
```

And in release scripts, we just update this text file and commit to repo, so the automatic build would be triggered on that repo too).

amarts:
Ok, for now, lets get this with changing the part  where it says 0.7.7 to 0.8.0 ?? (here and also in quick-start.md?)

aravindavk:
> Ok, for now, lets get this with changing the part where it says 0.7.7 to 0.8.0 ?? (here and also in quick-start.md?)

Ack. Short URL may look good in website. Like

```
curl -LO https://dl.kadalu.io/latest/kubectl-kadalu
```

That may need little work to set it up. I will change it to 0.8.0 in both the places

amarts:
It is ok to have a curl command inside of a another curl... Reason, we can use the similar method for moana, glusterfs etc from kadalu later... So documents can all look uniform. Just a thought.

amarts (approved):

PR #483, opened 2021-03-12
Replica client volfiles: set self-heal options to default

### LINKED ISSUE #477: Self-heal not triggered on files in 0.8.0
Looks like self-heal is not properly triggered on the files properly. It was working fine on 0.7.7 version!

amarts:
By this, even if self-heal daemon doesn't work, the client's access
itself can heal the file, thus providing users with an option to
continue, instead of being clueless what to do next.

Fixes: #477

### DIFF
--- a/templates/Replica2.client.vol.j2
+++ b/templates/Replica2.client.vol.j2
@@ -20,13 +20,12 @@ end-volume
 {% for i in range(dht_subvol|length) %}
 volume {{ volname }}-replica-{{ i }}
     type cluster/replicate
-    option data-self-heal off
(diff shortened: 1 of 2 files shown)


aravindavk (approved):

amarts:
even though it is provided as a valid option `data-self-heal open` is not working. Took almost 2 hrs to figure that out.

Best thing to handle this is by actually allowing to handle glusterfs_options, and taking log-level=DEBUG in there. That way, it becomes easier to look at the failures in CI logs and identify the issues.

Issue #471, opened 2021-03-09
[CI] - Add more IO tests to qualify the release


amarts:
It is good if we add more IO tests to the PR merge CIs (or at least release CIs). That we we can avoid bugs like #468 which happen after upgrade!

Recommend some automated containers which can be used for testing!

leelavg:
@amarts I've prepared an image (~75MB) consisting of Crefi, arequal and I believe this type of IO is good enough for a start.

If you are fine with the size and [dockerfile steps](https://thoughtexpo.com/setup-k3d-cluster-for-local-testing-or-development/#local-registry), i can give a try including that in CI.

amarts:
Sure. It seems like a good idea.

KyleSanderson:
Can this sample be added to CI https://github.com/kadalu/kadalu/issues/577#issuecomment-874283249

leelavg:
@KyleSanderson definitely, dbench seems to be a good candidate to be included in CI.

@vatsa287 see if you can pick this up or I'll start working on it after a week's time.

leelavg:
> Can this sample be added to CI [#577 (comment)](https://github.com/kadalu/kadalu/issues/577#issuecomment-874283249)

I revisited this and unfortunately current CI infra isn't suited for dbench, however we'll add more functional tests based on recent learnings around disk usage metrics

PR #484, opened 2021-03-12
Add minimal IO tests

leelavg:
- Deploy two IO pods
- Write IO sequentially from each pod
- Verify arequal after IO completion
- Ideally will extend CI run only by 2 minutes at max

Updates: https://github.com/kadalu/kadalu/issues/471

### DIFF
--- a/tests/io-app.yaml
+++ b/tests/io-app.yaml
@@ -1,0 +1,49 @@
+# -*- mode: yaml -*-
+---
+kind: PersistentVolumeClaim
+apiVersion: v1
+metadata:
+  name: pv-io
+spec:
+  storageClassName: kadalu.replica3
+  accessModes:
+    - ReadWriteMany
+  resources:
+    requests:
+      storage: 10Mi
+
+---
+apiVersion: apps/v1
+kind: Deployment
+metadata:
+  name: io-app
+spec:
+  replicas: 2
+  selector:
+    matchLabels:
+      app: io-app
+  template:
+    metadata:
+      labels:
+        app: io-app
+    spec:
+      containers:
+      - name: io-pod
+        image: docker.io/leelavg/dark-side:latest
+        imagePullPolicy: IfNotPresent
+        command: ['sh', '-c', 'echo "Ready!" && while true; do sleep 10; done;']
+        volumeMounts:
+        - mountPath: '/mnt/alpha'
+          name: csivol
+        livenessProbe:
+          exec:
+            command:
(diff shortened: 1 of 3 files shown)


leelavg:
Please build the image locally and push to docker hub once the Dockerfile is reviewed and I'll force push with updated image address.

tests/io-app.yaml
```diff
@@ -0,0 +1,49 @@
...
+    matchLabels:
+      app: io-app
+  template:
+    metadata:
+      labels:
+        app: io-app
+    spec:
+      containers:
+      - name: io-pod
+        image: docker.io/leelavg/dark-side:latest
```
amarts, line 32: you want to actually call it `dark-side`? or any particular name?

leelavg: Not at all 😅, please choose any name for the docker image before pushing to kadalu docker hub. 

You can also edit any labels in Dockerfile as well, just let me know of the edits I'll edit yaml and Dockerfile accordingly.

amarts: There are 2 thoughts I have!

1. Always build locally and use the same - May add just 3-4 mins extra in complete test time, which is OK. Benefit is, any changes in `tests/Dockerfile` also gets tested. (or add this build locally as a conditional step, only if the particular file changed in PR).

2. The PR merge job can look for any changes to `tests/Dockerfile` and then build an image (for now only `linux/amd64`) and push it to docker hub. As we already have secret and everything properly setup, this works best instead of manually pushing the image.

leelavg: Ack.

1. I can go with adding a conditional step for local build
2. This image is similar to sample-app in the sense gets updated very rare, will push conditionally based on file in PR

amarts:
I will do one thing! I will merge this PR as this is already an improvement. Please work on improvement as a separate PR.

amarts (approved):

leelavg:
Definitely, I've many to-do's wrt  tests, just noting it here to look back before raising a tracker issue:
- Remove unwanted o/p from Crefi in CI logs
- Increase size of PV (verify re-size), deploy one more pod from same deployment and run IO
- With IO running bring down one server and verify healinfo (need to activate cli-tests) and arequal
- Deciding on image (tests/Dockerfile) building to be part of workflow or not
- Run a pure gluster pod to verify external volume? (not immediate)
- Bring csi-sanity and e2e storage tests under CI (not immediate)

Issue #474, opened 2021-03-10
0.8.0 fails on AARCH64 (Pi 4 4GB)


DanielWinks:
When attempting to run 0.8.0 on AARCH64/ARM64, the operator fails to start, giving error message:

`ModuleNotFoundError: No module named 'xxhash._xxhash'`

Seems the AARCH64 build is missing the above Python module. 0.7.7 does not display this issue and works properly.

amarts:
Ack! 

This is my bad, there were some issues with `pynacl` (ie, cryptography pip install in arm/v7 tag). To figure that out, I had made both arm builds disabled. Added them back now, 0.8.1 should have the arm64 builds. (not yet ready with arm/v7).

Next time such things happen, it will have a pointer in Release notes.

amarts:
The issue of ARM64 is a major one we wanted to resolve, and a lot of work has gone in last 6 months for the same. With latest `devel` tag everything is working. Will make a release 0.8.6 soon which should have most of the things covered.

PR #488, opened 2021-03-14
build scripts: minor tweaks and enable arm64 and arm/v7

### LINKED ISSUE #486: Errors seen on Release CI
This is happening when I add `linux/arm64` platform to the build.

```
#39 3246.0     cc1: warning: command line option ‘-std=c+
#39 3246.0 [output clipped, log limit 1MiB reached]
#39 ERROR: executor failed running [/bin/sh -c python3 -m pip install grpcio googleapis-common-protos >/dev/null]: exit code: 1
------
 > [linux/arm64 builder 2/2] RUN python3 -m pip install grpcio googleapis-common-protos >/dev/null:
------
Dockerfile:5
--------------------
   3 |     ENV PATH="/kadalu/bin:/opt/bin:/opt/sbin:$PATH"
   4 |     
   5 | >>> RUN python3 -m pip install grpcio googleapis-common-protos >/dev/null
   6 |     
   7 |     FROM ubuntu:20.04 as prod
--------------------
error: failed to solve: rpc error: code = Unknown desc = executor failed running [/bin/sh -c python3 -m pip install grpcio googleapis-common-protos >/dev/null]: exit code: 1
Error: buildx call failed with: error: failed to solve: rpc error: code = Unknown desc = executor failed running [/bin/sh -c python3 -m pip install grpcio googleapis-common-protos >/dev/null]: exit code: 1
```

amarts:
* Install grpcio through `apt-get install python3-grpcio` (saves 20-30mins on arm builds)
* Install pyxattr through `apt-get install python3-pyxattr` (saves 20-30mins on arm builds)
  - previously `pynacl` (a dependency) used to take a lot of time with pip install.
* Install glustercli only in server builds.
* Install glusterfs with `make install-strip` to reduce the binary size of glusterfs

Tested on `arm` branch for release, so both `arm64` and `arm/v7` tag works

Fixes: #486 
Updates: #474 

### DIFF
--- a/server/Dockerfile
+++ b/server/Dockerfile
@@ -1,7 +1,6 @@
 FROM kadalu/builder:latest as builder
 
 ENV PATH="/kadalu/bin:/opt/bin:/opt/sbin:$PATH"
-RUN python3 -m pip install glustercli
 
 FROM ubuntu:20.04 as prod
 
@@ -15,9 +14,11 @@ COPY --from=builder /kadalu /kadalu
 
 RUN apt-get update -yq && \
     apt-get install -y --no-install-recommends python3 xfsprogs libtirpc3 sqlite3 && \
-    apt-get install -y --no-install-recommends python3-pyxattr && \
     apt-get -y clean && \
     rm -rf /var/lib/apt/lists/*
+
+# copy pyxattr installed in pypi_source image
+COPY --from=kadalu/pypi_source:latest /usr/local/lib/python3.8/dist-packages /kadalu/lib/python3.8/site-packages/
 
 RUN mkdir -p /kadalu/templates /kadalu/volfiles
 RUN mkdir -p /var/run/gluster /var/log/glusterfs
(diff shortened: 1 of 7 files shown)


server/Dockerfile
```diff
@@ -1,15 +1,21 @@
+FROM kadalu/builder:latest as builder
+
+ENV PATH="/kadalu/bin:/opt/bin:/opt/sbin:$PATH"
+RUN python3 -m pip install glustercli
```
aravindavk: This package was introduced when external Quotad was running gluster cli commands. Now this can be removed?

amarts: We don't have control on having 'simple-quota' on External glusterfs. So, It may be needed with kubect-quotad.

About having it in server image, I am not sure!

amarts: Ok, rechecked the code. We introduced glustercli in server image because we had quotad, which needed it. Now that its not there, I feel we can remove it.

amarts:
I see excessive logs related to grpcio! Will handle this differently!

leelavg:
Just checking my understanding, with this PR we are:

Gaining:
- No need to maintain arm branch
- Images of arm+amd will be built and pushed on every merge 
- Tweaks to gluster resulting in reduced build time

With tradeoffs:
- Build time of >2hrs on every merge

Did I miss anything?

amarts:
> With tradeoffs:
> 
> Build time of >2hrs on every merge

That is not true. For every merge, we will only build `linux/amd64` platform with `devel` tag. The ~1hr build time is only for release tagging.

amarts:
Merging, and triggering a pre-release build to test arm64 and arm/v7 builds.

PR #495, opened 2021-03-18
csi/nodeserver: Update csi-sanity for NodePublishVolume

vatsa287:
Updates: #494

Return with errmsg when volume_id, target_path or volume_capability is not provided.

Signed-off-by: Shree Vatsa N <vatsa@kadalu.io>

PR #496, opened 2021-03-18
Use bash 'test' command inside a conditional in CI step

leelavg:
- If bash 'test' ([]) command used outside of conditional
return code is set '1' on failure at the end
- Change to using 'test' inside a conditional statement


leelavg:
Change from 

```
# empty=''; [[ "$empty" == "value" ]] && echo at the end; echo $?
1
```
to
```
# empty=''; if [[ "$empty" == "value" ]]; then echo at the end; fi; echo $?
0
```

The CI was failing in https://github.com/kadalu/kadalu/pull/495 due to first form

amarts (approved):

Issue #494, opened 2021-03-18
[Tracker] Fix CSI sanity failures to conform to CSI Spec v1.2


leelavg:
# CSI Sanity

- Below are some good first issues for new contributors to get to know about KaDalu CSI Driver and CSI Spec

## Easy (good first issues):
- [x] Node Service NodePublishVolume [It] should fail when no volume id is provided
- [x] Node Service NodePublishVolume [It] should fail when no target path is provided
- [x] Node Service NodePublishVolume [It] should fail when no volume capability is provided
- [x] Node Service NodeUnpublishVolume [It] should fail when no volume id is provided
- [x] Controller Service [Controller Server] DeleteVolume [It] should fail when no volume id is provided
- [x] Controller Service [Controller Server] ValidateVolumeCapabilities [It] should fail when no volume id is provided
- [x] ExpandVolume [Controller Server] [It] should fail if no volume id is given
- [x] Controller Service [Controller Server] CreateVolume [It] should fail when no name is provided
- [x] Controller Service [Controller Server] CreateVolume [It] should fail when no volume capabilities are provided

## Easy to Medium (good first issues):
- [x] Controller Service [Controller Server] ValidateVolumeCapabilities [It] should fail when no volume capabilities are provided
- [x] Controller Service [Controller Server] ValidateVolumeCapabilities [It] should fail when the requested volume does not exist
- [x] Node Service NodeUnpublishVolume [It] should fail when no target path is provided

## Medium:
- [x] Controller Service [Controller Server] ListVolumes [It] should return appropriate values (no optional values added)
- [x] Controller Service [Controller Server] ListVolumes [It] should fail when an invalid starting_token is passed
- [x] Controller Service [Controller Server] ListVolumes [It] check the presence of new volumes and absence of deleted ones in the volume list
- [x] Controller Service [Controller Server] CreateVolume [It] should fail when requesting to create a volume with already existing name and different capacity.
- [x] Controller Service [Controller Server] ValidateVolumeCapabilities [It] should return appropriate values (no optional values added)
- [x] ExpandVolume [Controller Server] [It] should fail if no capacity range is given
- [x] ExpandVolume [Controller Server] [It] should work

#### For fixing above failure, the flow should be:
- Find failing service: `Node Service`
- Find failing RPC method: `NodePubhlishVolume`
- Actual test failure: `should fail when no volume id is provided`
- Refer CI logs corresponding to the test:
```
[...]

PR #503, opened 2021-03-30
csi: CreateVolume should fail with same name and different capacity

vatsa287:
Updates: #494

### DIFF
--- a/csi/controllerserver.py
+++ b/csi/controllerserver.py
@@ -53,6 +53,16 @@ class ControllerServer(csi_pb2_grpc.ControllerServicer):
             context.set_code(grpc.StatusCode.INVALID_ARGUMENT)
             return csi_pb2.CreateVolumeResponse()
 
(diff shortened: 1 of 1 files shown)


amarts:
Just a suggestion on CI.

Once this run completes, can you also make a change to CI tests on expected passes for csi-sanity ? That way, we know we are not regressing anymore, and with every PR, we have an idea of what the improvement is. Ie, every PR should have increment of 1 at least, so we know the tests are running as expected, instead of getting to logs.

amarts:
noticing some glusterfs crashes in replica2 setup. Considering we don't run any tests, remove it from CI. We need to test it separately, properly and enable it again.

vatsa287:
> noticing some glusterfs crashes in replica2 setup. Considering we don't run any tests, remove it from CI. We need to test it separately, properly and enable it again.

Still true..

PR #504, opened 2021-03-30
CI: tests - remove Replica2 from it

amarts:
Also increase the expected_pass for CSI-sanity to 20 (from 7).


amarts:
This needs some more debugging on glusterfs side, and will be enabled post that! Target 0.8.3

Important: (0.8.2 release notes should have a note about Replica2 not being tested).

PR #506, opened 2021-03-30
csi: change few test flows to pass csi-spec

amarts:
Updates: #494 

### DIFF
--- a/csi/controllerserver.py
+++ b/csi/controllerserver.py
@@ -7,9 +7,10 @@ import os
 import random
 import time
 
+import grpc
+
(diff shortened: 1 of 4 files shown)


amarts:
From the tests:
```
Ran 33 of 73 Specs in 10.035 seconds
SUCCESS! -- 33 Passed | 0 Failed | 0 Pending | 40 Skipped
Sanity [Pass %]: Expected: 33 and Actual: 33
```
So, at least 33 jobs should always pass in future to get CI pass. Which is good.

PR #511, opened 2021-04-11
Preparation for krew release

aravindavk:
Krew(https://krew.sigs.k8s.io) is a package manager for Kubectl Plugins.
Once the Kadalu manifest file is submitted to the krew_index repo,
users can install Kadalu as `kubectl krew install kadalu`.

It expects `zip` or `tar.gz` file available for download while
installing the plugin. With this PR, `tar.gz` file added to the
release and the krew manifest file is added to `extras/krew.yaml`

Following changes are required after each release to submit the
Kadalu manifest file to krew index.

* Update Version and download URL.
* Update sha256 of `kubectl-kadalu.tar.gz` file.

### DIFF
--- a/cli/Makefile
+++ b/cli/Makefile
@@ -26,6 +26,7 @@ release: gen-version
 	@cd build/src && zip -r ../${PROGNAME}.zip *
 	@echo '#!/usr/bin/env ${PYTHON}' | cat - build/${PROGNAME}.zip > build/${PROGNAME}
 	@chmod +x build/${PROGNAME}
+	@cd build && tar -czvf ${PROGNAME}.tar.gz ${PROGNAME}
 	@rm -rf build/src
 	@rm -f build/${PROGNAME}.zip
 	@echo "Single deployment file is ready: build/${PROGNAME}"
--- a/extras/krew.yaml
+++ b/extras/krew.yaml
@@ -1,0 +1,26 @@
+apiVersion: krew.googlecontainertools.github.com/v1alpha2
+kind: Plugin
+metadata:
+  name: kadalu
+spec:
+  version: "0.8.2"
+  homepage: https://github.com/kadalu/kadalu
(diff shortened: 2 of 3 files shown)


aravindavk:
I haven't opened PR in https://github.com/kubernetes-sigs/krew-index repo, I will do it after the 0.8.2 release since it expects `tar.gz` file of the binary.

cli/Makefile
```diff
@@ -26,6 +26,7 @@ release: gen-version
 	@cd build/src && zip -r ../${PROGNAME}.zip *
 	@echo '#!/usr/bin/env ${PYTHON}' | cat - build/${PROGNAME}.zip > build/${PROGNAME}
 	@chmod +x build/${PROGNAME}
+	@cd build && tar -czvf kubectl-kadalu.tar.gz kubectl-kadalu
```
leelavg: Small nit, I think it'll be good to use `${PROGNAME}` to be consistent?

extras/krew.yaml
```diff
@@ -0,0 +1,26 @@
+apiVersion: krew.googlecontainertools.github.com/v1alpha2
+kind: Plugin
+metadata:
+  name: kadalu
+spec:
+  version: "0.8.2"
```
leelavg, line 6: Just a thought (maybe in another PR), wouldn't it better to update `version` and `sha256` automatically as part of release and raise PR against krew-index?

I maybe wrong here, do we need to raise PR against krewindex after updating yaml as part of every release?

cli/Makefile
aravindavk: ack

extras/krew.yaml
aravindavk: Yes for now it is manual.

> I maybe wrong here, do we need to raise PR against krewindex after updating yaml as part of every release?

Yes by looking at the pull requests https://github.com/kubernetes-sigs/krew-index/pulls?q=is%3Apr+is%3Aclosed

amarts (approved):

PR #516, opened 2021-04-17
doc: fix storage.version to v1alpha instead of v1

amarts:


aravindavk:
Looks good. Any issue observed with version `v1`?

amarts:
> Looks good. Any issue observed with version `v1`?

A user reported that with v1 there was an error, and with v1alpha it worked fine. Hence the change.

aravindavk (approved):

Issue #538, opened 2021-05-11
version in helm chart not matches with release version


ben-wangz:
i found the version in helm chart not matches with release version.

* running the command below: 

``` 
   helm install kadalu https://github.com/kadalu/kadalu/releases/download/0.8.2/kadalu-helm-chart.tgz --dry-run
```

* the output shows 
    1. image: docker.io/kadalu/kadalu-operator:devel
    2. the env KADALU_VERSION is set to devel too

* i have checked [helm/kadalu/templates/deployment.yaml](/kadalu/kadalu/blob/524dd2317ddbd4c2f4d52acd19c4dd6daa861000/helm/kadalu/templates/deployment.yaml). the image and the env KADALU_VERSION won't be changed (or i may have missed something)
* will it be updated in next version in plan? may i submit a PR if not planed?

```
NAME: kadalu
LAST DEPLOYED: Tue May 11 12:10:36 2021
NAMESPACE: default
STATUS: pending-install
REVISION: 1
TEST SUITE: None
HOOKS:
MANIFEST:
---
# Source: kadalu/templates/serviceaccount.yaml
apiVersion: v1
kind: ServiceAccount
metadata:
  name: kadalu-operator
  namespace: default
---
# Source: kadalu/templates/serviceaccount.yaml
kind: ServiceAccount
apiVersion: v1
metadata:
  name: kadalu-csi-nodeplugin
  namespace: default
---
# Source: kadalu/templates/serviceaccount.yaml
kind: ServiceAccount
apiVersion: v1
metadata:
  name: kadalu-csi-provisioner
  namespace: default
---
# Source: kadalu/templates/serviceaccount.yaml
kind: ServiceAccount
apiVersion: v1
metadata:
  name: kadalu-server-sa
  namespace: default
---
# Source: kadalu/templates/resources.yaml
apiVersion: apiextensions.k8s.io/v1
kind: CustomResourceDefinition
metadata:
  name: kadalustorages.kadalu-operator.storage
  namespace: default
spec:
  group: kadalu-operator.storage
  names:
    kind: KadaluStorage
    listKind: KadaluStorageList
    plural: kadalustorages
    singular: kadalustorage
  scope: Namespaced
  versions:
    - name: v1alpha1
      storage: true
      served: true
      schema:
        openAPIV3Schema:
          type: object
          properties:
            spec:
              type: object
              properties:
                type:
                  type: string
                storage:
                  type: array
                  items:
                    type: object
                    properties:
                      node:
[...]

PR #539, opened 2021-05-11
kadalu-operator image version and KADALU_VERSION, see also #538

ben-wangz:
1. lightly change: use the value in .Chart.Version which has already been auto replaced by the build scripts
2. "devel" is reserved as the default
3. i did not found any else bind with "devel" in helm
4. NOTE
    * sorry for not fully tested by scripts in tests/, because i cannot turn minikube up with my local envrionment(apple m1 chip)
    * but i have tested with
        1. ``` export KADALU_VERSION=0.8.2-b && make helm-chart ```
        2. ``` helm install kadalu kadalu-helm-chart.tgz --dry-run ```
        3. and checked the diff with 0.8.2/kadalu-helm-chart.tgz
5. any advice or idea is welcome

### DIFF
--- a/helm/kadalu/templates/deployment.yaml
+++ b/helm/kadalu/templates/deployment.yaml
@@ -23,7 +23,7 @@ spec:
           securityContext:
             capabilities: {}
             privileged: true
-          image: docker.io/kadalu/kadalu-operator:devel
+          image: "docker.io/kadalu/kadalu-operator:{{ .Chart.Version | default "devel" }}"
           imagePullPolicy: IfNotPresent
           env:
             - name: WATCH_NAMESPACE
@@ -39,7 +39,7 @@ spec:
             - name: DOCKER_USER
               value: "kadalu"
             - name: KADALU_VERSION
-              value: "devel"
+              value: "{{ .Chart.Version | default "devel" }}"
             - name: KADALU_NAMESPACE
               value: "{{ .Release.Namespace }}"
             - name: KUBELET_DIR


amarts (approved):

leelavg:
Thanks for PR, with current state it's tricky to get everything right. Please refer below:

1. Yes, for releases the PR will work, no doubt in that
2. If I want to clone the repo and install helm chart it'll fail because `version` is set to `0.0.0-0` and pulling operator image will fail and we can't set `version` to `devel` directly due to strict semver2 check
3. CI also would fail because pre-made manifests will have `devel` and helm generated manifest will have `0.0.0-0`

We are in this state due to some earlier agreed steps (but thanks for pointing out the mistake), the easiest way would be to figure out if helm can access environment variable or not.

ben-wangz:
* thx a lot @leelavg 
* do i need to change "devel" in manifests/kadalu-operator*.yaml to "0.0.0-0" ? or just keep this state of PR and waiting for the update of tests?
* not fully understand the second point, but i guess this won't block the merge according to the [test log](https://github.com/kadalu/kadalu/pull/539/checks?check_run_id=2553515513)

leelavg:
> * not fully understand the second point, but i guess this won't block the merge according to the [test log](https://github.com/kadalu/kadalu/pull/539/checks?check_run_id=2553515513)

A change to helm template should tick all below cases:
- If someone clone the repo and install operator via helm it should install `devel` version
- In CI, helm generated manifest (sink) should match manifests generated by `make gen-manifest` (source)
- In release helm should install correct version as specified in `.Chart.Version`

Can you please go through below and change accordingly after you understand above three points?
``` diff
diff --git a/helm/kadalu/templates/deployment.yaml b/helm/kadalu/templates/deployment.yaml
index f6bb28f..25f00fe 100644
--- a/helm/kadalu/templates/deployment.yaml
+++ b/helm/kadalu/templates/deployment.yaml
@@ -23,7 +23,7 @@ spec:
           securityContext:
             capabilities: {}
             privileged: true
-          image: docker.io/kadalu/kadalu-operator:devel
+          image: docker.io/kadalu/kadalu-operator:{{- if eq .Chart.Version "0.0.0-0" }}devel{{- else}}{{.Chart.Version}}{{- end }}
           imagePullPolicy: IfNotPresent
           env:
             - name: WATCH_NAMESPACE
@@ -39,7 +39,7 @@ spec:
             - name: DOCKER_USER
               value: "kadalu"
             - name: KADALU_VERSION
-              value: "devel"
+              value: "{{- if eq .Chart.Version "0.0.0-0" }}devel{{- else}}{{.Chart.Version}}{{- end }}"
             - name: KADALU_NAMESPACE
               value: "{{ .Release.Namespace }}"
             - name: KUBELET_DIR
```
Edit: Changed single quote to double quote.

ben-wangz:
After reading the 'make gen-manifest' logic, it's clear to me now.

Having tried to create a variable in values.yaml and replace "0.0.0-0" just like kadalu/Chart.yaml
But then I realized another difficulty: helm3 won't accept `chart.metadata.version = devel`
In other words, we cannot achieve: "If someone clone the repo and install operator via helm it should install devel version"

Then "some earlier agreed steps" made a trick.

---

There's no good solutions right now I think. I will keep exploring the codes and try to found out a solution.
This PR can be closed at any time. Because these codes to merge is low-value.

---

thx again, i have learned a lot.

leelavg:
@ben-wangz glad that my pointers made sense now.

No need to close PR, it's a genuine issue to be fixed. Above diff o/p is the solution.

If I raise PR then I'll be incorporating above diff itself, since you found the issue and possible solution, you can amend this PR and push/force-push the commit.

Then the CI will pass and PR will be merged.

amarts:
> This PR can be closed at any time. Because these codes to merge is low-value.

Two reasons why we would appreciate your contribution here to reach till the PR is merged.

1. This solves a genuine bug.
2. Project gets one more contributor in its Contributor list :-)

Hence, whenever you get some free time, it would be very great if you can do the changes @leelavg mentioned, so we can pass the tests, and also merge the PR.

leelavg:
@ben-wangz hope things are going well on your end. This is too much of an important PR which shouldn't be hanging around and so I had closed it crediting you in #544

Please feel free to comment and pick up any other issues if interested.

PR #541, opened 2021-05-11
Add support for archiving pvc

### LINKED ISSUE #535: have an option to archive volumes on delete
Current behaviour is to delete the data whenever a PVC is deleted.  Would be useful to have an option to retain the data when the PVC gets deleted.   Similar to the behaviour of nfs-subdir-external-provisioner, which (if configured) renames the volume to archived-<volname> if the PVC gets deleted.

vatsa287:
This PR adds support for archiving pvc.

- New 'pvReclaimPolicy' option in the storage config which takes any one of ["delete", "archive"] values and "delete" is default.
- When a PVC is deleted with "archive" option, the PVC to be deleted is archived ( Renamed to archived-pvc-* )

Fixes: #535

### DIFF
--- a/csi/controllerserver.py
+++ b/csi/controllerserver.py
@@ -301,7 +301,7 @@ class ControllerServer(csi_pb2_grpc.ControllerServicer):
 
         delete_volume(request.volume_id)
         logging.info(logf(
-            "Volume deleted",
+            "Volume deleted/archived",
             name=request.volume_id,
             duration_seconds=time.time() - start_time
         ))
--- a/csi/controllerserver.py
+++ b/csi/controllerserver.py
@@ -301,7 +301,7 @@ class ControllerServer(csi_pb2_grpc.ControllerServicer):
 
         delete_volume(request.volume_id)
         logging.info(logf(
-            "Volume deleted/archived",
+            "Delete Volume response completed",
             name=request.volume_id,
             duration_seconds=time.time() - start_time
         ))
--- a/csi/volumeutils.py
+++ b/csi/volumeutils.py
@@ -589,6 +589,7 @@ def update_pv_metadata(hostvol_mnt, pvpath, expansion_requested_pvsize):
 
     logging.debug(logf(
         "Metadata updated",
+        metadata_file=info_file_path
     ))
 
 
@@ -615,40 +616,51 @@ def delete_volume(volname):
     retry_errors(os.statvfs, [os.path.join(HOSTVOL_MOUNTDIR, vol.hostvol)],
                  [ENOTCONN])
 
-    filename = vol.hostvol + ".info"
-    with open(os.path.join(VOLINFO_DIR, filename)) as info_file:
+    storage_filename = vol.hostvol + ".info"
+    with open(os.path.join(VOLINFO_DIR, storage_filename)) as info_file:
         storage_data = json.load(info_file)
 
     onDelete = storage_data.get("onDelete", "delete")
 
     volpath = os.path.join(HOSTVOL_MOUNTDIR, vol.hostvol, vol.volpath)
 
-    # Get brick data
-    info_file_path = os.path.join(HOSTVOL_MOUNTDIR, vol.hostvol, "info", vol.volpath)
-    with open(info_file_path + ".json", "r") as info_file:
-        brick_data = json.load(info_file)
-
-    path_prefix = brick_data["path_prefix"]
+    # # Get brick data
+    # info_file_path = os.path.join(HOSTVOL_MOUNTDIR, vol.hostvol, "info", vol.volpath)
+    # with open(info_file_path + ".json", "r") as info_file:
+    #     brick_data = json.load(info_file)
+    # path_prefix = brick_data["path_prefix"]
 
     logging.info(logf(
         "onDelete data",
         hostvolume=vol.hostvol,
         onDelete=onDelete,
         volpath=volpath,
-        destination=os.path.join(HOSTVOL_MOUNTDIR, vol.hostvol, vol.volpath),
-        path_prefix=path_prefix
+        #destination=os.path.join(HOSTVOL_MOUNTDIR, vol.hostvol, vol.volpath),
+        #path_prefix=path_prefix,
     ))
 
     if onDelete == "archive":
+
         old_volname = vol.volname
         vol.volname = "archived-" + vol.volname
+        path_prefix = os.path.dirname(vol.volpath)
         vol.volpath = os.path.join(path_prefix, vol.volname)
 
-        # Rename directory
+        # Rename directory & files that are to be archived
         try:
+
+            # Brick/PVC
             os.rename(
                 os.path.join(HOSTVOL_MOUNTDIR, vol.hostvol, path_prefix, old_volname),
                 os.path.join(HOSTVOL_MOUNTDIR, vol.hostvol, path_prefix, vol.volname)
+            )
+
+            # Info-File
+            old_info_file_name = old_volname + ".json"
+            info_file_name = vol.volname + ".json"
+            os.rename(
+                os.path.join(HOSTVOL_MOUNTDIR, vol.hostvol, "info", path_prefix, old_info_file_name),
+                os.path.join(HOSTVOL_MOUNTDIR, vol.hostvol, "info", path_prefix, info_file_name)
             )
 
             logging.info(logf(
--- a/csi/volumeutils.py
+++ b/csi/volumeutils.py
@@ -666,9 +666,9 @@ def delete_volume(volname):
             )
 
             logging.info(logf(
-                "Volume has been archived",
+                "Volume archived",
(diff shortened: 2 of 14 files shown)


operator/main.py
```diff
@@ -581,6 +587,14 @@ def handle_modified(core_v1_client, obj):
     # Get volume-id from config map
     obj["spec"]["volume_id"] = cfgmap["volume_id"]
 
+    logging.logf(info(
```
amarts: this looks to be not right... isn't this `logging.info(logf())` ?

csi/volumeutils.py
```diff
@@ -610,11 +610,65 @@ def delete_volume(volname):
...
+    onDelete = storage_data.get("onDelete", "delete")
+
     volpath = os.path.join(HOSTVOL_MOUNTDIR, vol.hostvol, vol.volpath)
+
+    # Get brick data
+    info_file_path = os.path.join(HOSTVOL_MOUNTDIR, vol.hostvol, "info", vol.volpath)
+    with open(info_file_path + ".json", "r") as info_file:
+        brick_data = json.load(info_file)
+
+    path_prefix = brick_data["path_prefix"]
```
amarts: I see usage of `path_prefix` only in onDelete == "archive", can get inside that section.

templates/operator.yaml.j2
```diff
@@ -95,6 +95,8 @@ spec:
                       type: integer
                 type:
                   type: string
+                onDelete:
```
leelavg: How about using `delete` as default here and remove `onDelete: delete` from all storage-classes?
Then we just need to disregard value for this key only for `external non-native storage`, thoughts?

PS: Don't forget update in helm as well :)

csi/volumeutils.py
```diff
@@ -610,11 +610,65 @@ def delete_volume(volname):
...
+    logging.info(logf(
+        "onDelete data",
+        hostvolume=vol.hostvol,
+        onDelete=onDelete,
+        volpath=volpath,
+        destination=os.path.join(HOSTVOL_MOUNTDIR, vol.hostvol, vol.volpath),
+        path_prefix=path_prefix
+    ))
+
+    if onDelete == "archive":
```
leelavg: Do you feel skipping this step https://github.com/kadalu/kadalu/blob/devel/csi/quota-crawler.sh#L22 for `archive` pvcs is good ?

csi/volumeutils.py
leelavg: @amarts It's being used in logging function at L639 :)

leelavg (commented):
Some general observations.

- Would it be possible to update docs at https://kadalu.io/docs/k8s-storage/devel/storage-config-options/ regarding delete option?
- Would it be possible to add/edit some tests (only if you think it's needed)?

templates/operator.yaml.j2
vatsa287: Yes its default, I have added it has a reference. like onDelete = obj["spec"].get("onDelete", "delete")

vatsa287: I'll update the docs once I complete this PR properly.

leelavg: > Yes its default, I have added it has a reference. like onDelete = obj["spec"].get("onDelete", "delete")

ack, I meant using default in crd itself like below, disregard if you mean the same:
``` yaml
onDelete:
  type: string
  default: delete
```

operator/main.py
vatsa287: Yeah just added for debug

templates/operator.yaml.j2
vatsa287: cool will do it, I had meant the other one

amarts (approved):

aravindavk:
Option name `onDelete` is confusing that it is Storage(Host Volume) delete or PV delete. Can we change it to `onPvDelete` or something?

csi/controllerserver.py
```diff
@@ -301,7 +301,7 @@ def DeleteVolume(self, request, context):
 
         delete_volume(request.volume_id)
         logging.info(logf(
-            "Volume deleted",
+            "Volume deleted/archived",
```
aravindavk: Add exact message so that from the log it will be easy to identify it is deleted or archived. Also add archived name as `archive_name="archived-" + request.volume_id`

aravindavk:
We can follow similar terminologies as available in Storage Class. `pvReclaimPolicy=delete|retain`. In future when we add support for `reclaimPolicy` for Storage Class then that will get high priority than the default one set while creating Kadalu Storage. For example, if reclaim_policy_from_storage_class is None then use reclaim_policy_from_storage_info.
 
https://kubernetes.io/docs/tasks/administer-cluster/change-pv-reclaim-policy/

> the default reclaim policy is "Delete". This means that a dynamically provisioned volume is automatically deleted when a user deletes the corresponding PersistentVolumeClaim.

> With the "Retain" policy, if a user deletes a PersistentVolumeClaim, the corresponding PersistentVolume is not be deleted. Instead, it is moved to the Released phase, where all of its data can be manually recovered.

csi/controllerserver.py
```diff
@@ -301,7 +301,7 @@ def DeleteVolume(self, request, context):
 
         delete_volume(request.volume_id)
         logging.info(logf(
-            "Volume deleted",
+            "Delete Volume response completed",
```
vatsa287, line 304: @aravindavk Checking for hostvol & then pvReclaimPolicy type would require search_volume() again, so specific logs will be made from volumeutils.py itself

csi/volumeutils.py
```diff
@@ -696,7 +696,7 @@ def delete_volume(volname):
             error=err,
         ))
 
-    logging.debug(logf(
+    logging.info(logf(
```
vatsa287, line 684: So when deleted this would be triggered

csi/volumeutils.py
vatsa287: Since the data is still preserved, should the stats of Hosting Volume be updated. I think when the archived data is manually removed as in reclaimPolicy: retain, the occupied size should be subtracted.

vatsa287:
### Test Run
```bash
vatsa@vatsa-VirtualBox:~/kadalu$ sudo kubectl logs kadalu-csi-provisioner-0 kadalu-provisioner -nkadalu
[2021-05-13 14:41:42,545] INFO [kadalulib - 339:start_all] - Started Process	 name=csi
[2021-05-13 14:41:42,549] INFO [kadalulib - 339:start_all] - Started Process	 name=metrics
[2021-05-13 14:41:42,681] INFO [kadalulib - 339:start_all] - Started Process	 name=quota
Starting the quota crawler script
No PVC yet, continuing to watch...
[2021-05-13 14:41:51,873] INFO [exporter - 85:<module>] - Started Kadalu Storage CSI Metrics exporter.	 port=8000
[2021-05-13 14:41:56,789] INFO [main - 41:mount_storage] - Volume is mounted successfully	 hvol=storage-pool-1
[2021-05-13 14:41:56,800] INFO [main - 61:main] - Server started
[2021-05-13 14:42:28,638] INFO [controllerserver - 265:CreateVolume] - Volume created	 name=pvc-0cd66ea3-c3e0-45b9-b0b6-2648d04bd254 size=104857600 hostvol=storage-pool-1 pvtype=subvol volpath=subvol/b6/7c/pvc-0cd66ea3-c3e0-45b9-b0b6-2648d04bd254 duration_seconds=27.3537437915802
[2021-05-13 14:42:43,089] INFO [volumeutils - 653:delete_volume] - Volume archived	 old_volname=pvc-0cd66ea3-c3e0-45b9-b0b6-2648d04bd254 new_archived_volname=archived-pvc-0cd66ea3-c3e0-45b9-b0b6-2648d04bd254 volpath=subvol/b6/7c/archived-pvc-0cd66ea3-c3e0-45b9-b0b6-2648d04bd254
[2021-05-13 14:42:43,089] INFO [controllerserver - 303:DeleteVolume] - Delete Volume response completed	 name=pvc-0cd66ea3-c3e0-45b9-b0b6-2648d04bd254 duration_seconds=0.0211942195892334
setfattr: /mnt/storage-pool-1/subvol/b6/7c/archived-pvc-0cd66ea3-c3e0-45b9-b0b6-2648d04bd254: Operation not permitted
``` 
setfattr warning on renaming, idk if it's only on my device or not !? The files/directory are being renamed successfully.

csi/controllerserver.py
aravindavk: ok

csi/volumeutils.py
leelavg: Ack, I see little overhead in updating the size even for archived pvcs. Above can be disregarded then.

leelavg (approved):
LGTM. Seems we're getting `setfattr` errors lately, need to be looked in separate PR?

On other note, make sure kadalu cli is also updated in another PR.

aravindavk (approved):

PR #549, opened 2021-05-18
Fix regression due to removing `namespace` in helm CRD

leelavg:
Ref: #534


leelavg:
I don't think code change in #534 worth this regression. Just an opinion.

amarts (approved):
It was my mistake to not check all the details. Thanks @leelavg :+1:

PR #550, opened 2021-05-18
allow modification of Replica1 type to Replica2 or Replica3

amarts:

### DIFF
--- a/csi/main.py
+++ b/csi/main.py
@@ -4,6 +4,8 @@ Starting point of CSI driver GRP server
 import logging
 import os
 import time
+import signal
+
 from concurrent import futures
 
 import grpc
@@ -14,7 +16,7 @@ from identityserver import IdentityServer
 from kadalulib import CommandException, logf, logging_setup
 from nodeserver import NodeServer
 from volumeutils import (HOSTVOL_MOUNTDIR, get_pv_hosting_volumes,
-                         mount_glusterfs)
+                         mount_glusterfs, reload_glusterfs)
 
 _ONE_DAY_IN_SECONDS = 60 * 60 * 24
 
@@ -41,6 +43,21 @@ def mount_storage():
         logging.info(logf("Volume is mounted successfully", hvol=hvol))
     return
 
+def reconfigure_mounts():
+    """
+    Reconfigure the mounts by regenerating the volfiles.
+    """
+    host_volumes = get_pv_hosting_volumes({})
+    for volume in host_volumes:
+        if volume["type"] == "External":
+            # Need to skip remount external
+            continue
+        hvol = volume["name"]
+        try:
+            if reload_glusterfs(volume):
+                logging.info(logf("Volume reloaded successfully", hvol=hvol))
+        except CommandException:
+            logging.error(logf("Unable to mount volume", hvol=hvol))
 
 def main():
     """
@@ -51,6 +68,8 @@ def main():
 
     # If Provisioner pod reboots, mount volumes if they exist before reboot
     mount_storage()
+
+    signal.signal(signal.SIGHUP, reconfigure_mounts)
 
     server = grpc.server(futures.ThreadPoolExecutor(max_workers=10))
     csi_pb2_grpc.add_ControllerServicer_to_server(ControllerServer(), server)
--- a/csi/main.py
+++ b/csi/main.py
@@ -43,7 +43,8 @@ def mount_storage():
         logging.info(logf("Volume is mounted successfully", hvol=hvol))
     return
 
-def reconfigure_mounts():
+
+def reconfigure_mounts(_signum, _frame):
     """
     Reconfigure the mounts by regenerating the volfiles.
     """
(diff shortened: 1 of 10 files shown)


amarts:
This should be enough to support 'distribute' to 'replicate' type of the volume. But keeping it as draft as not sure what are the checks/validations we need to handle to prevent other disastrous situation where users can spoil an healthy volume.

Also, need a mechanism to regenerate the volume files for the volume which changed, so running volumes in nodeplugin and other places can properly work.

leelavg:
As discussed earlier on slack about changing Replica1 to ~other~ Replica2 before a migration, just curious are you aiming to facilitate replace-brick related to #537 with this PR?

aravindavk:
> As discussed earlier on slack about changing Replica1 to ~other~ Replica2 before a migration, just curious are you aiming to facilitate replace-brick related to #537 with this PR?

This is add brick + change of Volume type.

leelavg:
> This is add brick + change of Volume type.

Ack, what I was aiming at is, if we are able to change volume type if there's a possibility of changing back from Replica2 to Replica1 I can see a possibility of using it for migrating bricks.

Not to hijack this PR, I'll raise an RFE and then discuss on it's feasibility.

amarts:
If someone has kadalu setup handy, can this branch be tested? 

How to test ?

Create a storage pool with just one brick, now modify the same storage pool to be of Replica3 type. The major check to do is to check if the mounts in csi pods are now consuming new volfile.

csi/main.py
```diff
@@ -41,6 +43,21 @@ def mount_storage():
         logging.info(logf("Volume is mounted successfully", hvol=hvol))
     return
 
+def reconfigure_mounts():
```
leelavg: Single handler expects a func should accept two args, can you pls change the signature to `reconfigure_mounts(*args)`?

csi/volumeutils.py
```diff
@@ -824,6 +826,16 @@ def generate_client_volfile(volname):
     with open(info_file_path) as info_file:
         data = json.load(info_file)
 
+    if not VOL_DATA.get(volname, None):
```
leelavg, line 833: If possible can you please add some comments for this snippet?

csi/watch-vol-changes.sh
```diff
@@ -0,0 +1,11 @@
+#!/bin/bash
+
+CONFIG_MAP_INFO_DIR="/var/lib/gluster"
+
+CSI_PROCESS_ID=$(ps aux|grep python | awk '{ print $2}' | xargs);
```
leelavg: Better grep for only `[m]ain.py`, `start.py` doesn't know how to handle `SIGHUP` and it'll restart whole pod.
Disregard if the intention is to restart all CSI pods.

amarts: You may be right. Intention is to only send signal to `main.py`. Still testing. My local setup is still broken, and wanted to see what makes sense. Expect changes here.

hunter86bg:
Can I use something like

 $ kubectl kadalu storage-add storage-pool-1 --type Replica3 \
     --path kube2:/bricks/brick2/brick --path kube3:/bricks/brick3/brick

to test replica1 to replica3 ?

csi/main.py
amarts: Ack

leelavg:
> $ kubectl kadalu storage-add storage-pool-1 --type Replica3
> --path kube2:/bricks/brick2/brick --path kube3:/bricks/brick3/brick

@hunter86bg thanks for your interest. For testing the change first you need to pull changes and build containers  (`make build-containers`) locally if not doing any hacks.

Atm, Amar is making use of CI to come to a final conclusion so I'd suggest pls wait until code path is fixed.

amarts:
Ok, my observation is that inotify is not working on configmap changes. So, need some other strategy.

amarts:
As SIGHUP was not working properly, I took the little 'not-so-efficient' but works properly strategy of checking for hash value of the configmap file whenever a 'mount_glusterfs()' is called. This is not efficient, but adding few milliseconds to every PV create, and every management operation is OK IMO to be correct.

leelavg:
After a bit of search about any existing issues with configMap and inotify, found [this](https://github.com/kubernetes/minikube/issues/1551#issuecomment-494957948)

amarts:
Thanks for the link @leelavg ! It made sense... I didn't suspect minikube earlier. Good to know.

Regardless, the latest implementation is ready to try out separately! I am intending to keep both SIGHUP and reload volfile during every mount call in code for another few months so we can see which is working better.

csi/main.py
```diff
@@ -42,6 +44,23 @@ def mount_storage():
...
+    for volume in host_volumes:
+        if volume["type"] == "External":
+            # Need to skip remount external
+            continue
+        hvol = volume["name"]
+        try:
+            if reload_glusterfs(volume):
+                logging.info(logf("Volume reloaded successfully", hvol=hvol))
+        except CommandException:
+            logging.error(logf("Unable to mount volume", hvol=hvol))
```
aravindavk, line 61: Wrong error message

amarts: Resolved in #555

amarts:
Taken up in #555

PR #551, opened 2021-05-19
[CI] Use abs path and github $HOME env var in release workflow

### LINKED ISSUE #546: Uploading artifacts in release workflow
- [Failure](https://github.com/kadalu/kadalu/runs/2486683225?check_suite_focus=true#step:12:8) in latest (0.8.2) release while uploading artefacts is due to unavailability of `output/downloads` folder which was created in [previous step](https://github.com/kadalu/kadalu/blob/devel/.github/workflows/on-release-tag.yml#L66)
- The folder should persist between steps and need to ~be investigated further~ look for an alternative if that's not the case.

leelavg:
Fixes: #546

### DIFF
--- a/.github/workflows/on-pr-submit.yml
+++ b/.github/workflows/on-pr-submit.yml
@@ -13,8 +13,27 @@ on:
 # Visit: https://github.blog/changelog/2021-02-08-github-actions-skip-pull-request-and-push-workflows-with-skip-ci/
 
 jobs:
-  test-dir:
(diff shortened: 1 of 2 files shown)


leelavg:
Persistence of folder contents can be viewed from [this run](https://github.com/kadalu/kadalu/runs/2618018547?check_suite_focus=true)

amarts (approved):

PR #555, opened 2021-05-22
Volume management

amarts:
* Support Replica1 -> Something else (best done to Replica2/Replica3)
* Support brick / server migration with decommissioned brick.
* Allow volume config changes to be picked up by clients as and when the storage config changes happen.

### DIFF
--- a/csi/controllerserver.py
+++ b/csi/controllerserver.py
@@ -469,6 +469,20 @@ class ControllerServer(csi_pb2_grpc.ControllerServicer):
             context.set_code(grpc.StatusCode.INVALID_ARGUMENT)
             return csi_pb2.ControllerExpandVolumeResponse()
 
+        if not request.capacity_range:
+            errmsg = "Capacity Range is empty and must be provided"
+            logging.error(errmsg)
+            context.set_details(errmsg)
+            context.set_code(grpc.StatusCode.INVALID_ARGUMENT)
+            return csi_pb2.ControllerExpandVolumeResponse()
+
+        if not request.capacity_range.required_bytes:
+            errmsg = "Capacity Range is empty and must be provided"
+            logging.error(errmsg)
+            context.set_details(errmsg)
+            context.set_code(grpc.StatusCode.INVALID_ARGUMENT)
+            return csi_pb2.ControllerExpandVolumeResponse()
+
         expansion_requested_pvsize = request.capacity_range.required_bytes
 
         # Get existing volume
--- a/csi/volumeutils.py
+++ b/csi/volumeutils.py
@@ -837,27 +837,34 @@ def generate_client_volfile(volname):
 
     # Tricky to get this right, but this solves all the elements of distribute in code :-)
     data['dht_subvol'] = []
+    decommissioned = []
     if data["type"] == "Replica1":
         for brick in data["bricks"]:
-            data["dht_subvol"].append("%s-client-%d" % (data["volname"], brick["brick_index"]))
+            brick_name = "%s-client-%d" % (data["volname"], brick["brick_index"])
+            data["dht_subvol"].append(brick_name)
+            if brick.get("decommissioned", "") != "":
+                decommissioned.append(brick_name)
     else:
         count = 3
         if data["type"] == "Replica2":
             count = 2
 
-        data["subvol_bricks_count"] = count
         if data["type"] == "Disperse":
-            data["subvol_bricks_count"] = data["disperse"]["data"] + \
-              data["disperse"]["redundancy"]
+            count = data["disperse"]["data"] + data["disperse"]["redundancy"]
             data["disperse_redundancy"] = data["disperse"]["redundancy"]
 
-        for i in range(0, int(len(data["bricks"]) / data["subvol_bricks_count"])):
-            data["dht_subvol"].append("%s-%s-%d" % (
+        data["subvol_bricks_count"] = count
+        for i in range(0, int(len(data["bricks"]) / count)):
+            brick_name = "%s-%s-%d" % (
                 data["volname"],
                 "disperse" if data["type"] == "Disperse" else "replica",
                 i
-            ))
+            )
+            data["dht_subvol"].append(brick_name)
+            if data["bricks"][(i * count)].get("decommissioned", "") != "":
+                decommissioned.append(brick_name)
 
+    data['decommissioned'] = "" if decommissioned == [] else ",".join(decommissioned)
     template_file_path = os.path.join(
         TEMPLATES_DIR,
         "%s.client.vol.j2" % data["type"]
--- a/csi/volumeutils.py
+++ b/csi/volumeutils.py
@@ -894,17 +894,21 @@ def reload_glusterfs(volume):
     with mount_lock:
         if not generate_client_volfile(volname):
             return False
-        pid = VOL_DATA[volname]["pid"]
         # TODO: ideally, keep the pid in structure for easier access
-        #cmd = ["kill", "-HUP", str(pid)]
-        cmd = ["ps", "aux", "|",
-               "grep", "glusterfs", "|",
-               "grep", "%s.client" % volname, "|",
-               "awk", "'{ print $2 }'", "|",
-               "xargs", "kill", "-HUP" ]
+        # pid = VOL_DATA[volname]["pid"]
+        # cmd = ["kill", "-HUP", str(pid)]
+        cmd = ["ps", "--no-header", "-ww", "-o", "pid,command", "-C", "glusterfs"]
 
         try:
-            execute(*cmd)
+            out, err, _ = execute(*cmd)
+            for line in out.split("\n"):
+                parts = line.split()
+                pid = parts[0]
+                for part in parts:
+                    if part.startswith("--volume-id="):
+                        if part.split("=")[-1] == volname:
+                            cmd1 = [ "kill", "-HUP", pid ]
+                            execute(*cmd1)
         except CommandException as err:
             logging.error(logf(
                 "error to execute command",
--- a/csi/volumeutils.py
+++ b/csi/volumeutils.py
@@ -26,6 +26,10 @@ VOLFILES_DIR = "/kadalu/volfiles"
 TEMPLATES_DIR = "/kadalu/templates"
 VOLINFO_DIR = "/var/lib/gluster"
 
+# This variable contains in-memory map of all glusterfs processes (hash of volfile and pid)
+# Used while sending SIGHUP during any modifcation to storage config
+VOL_DATA = {}
+
 statfile_lock = threading.Lock()    # noqa # pylint: disable=invalid-name
 mount_lock = threading.Lock()    # noqa # pylint: disable=invalid-name
 
@@ -640,9 +644,9 @@ def delete_volume(volname):
             info_file_name = vol.volname + ".json"
             os.rename(
                 os.path.join(HOSTVOL_MOUNTDIR, vol.hostvol, "info",
-                    path_prefix, old_info_file_name),
+                             path_prefix, old_info_file_name),
                 os.path.join(HOSTVOL_MOUNTDIR, vol.hostvol, "info",
-                    path_prefix, info_file_name)
+                             path_prefix, info_file_name)
             )
 
             logging.info(logf(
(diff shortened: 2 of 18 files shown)


amarts:
can't pass pid directly to cmd


```
[2021-05-23 01:32:47,215] ERROR [_server - 445:_call_behavior] - Exception calling application: expected str, bytes or os.PathLike object, not int
  File "/kadalu/lib/python3.8/site-packages/grpc/_server.py", line 435, in _call_behavior
  File "/kadalu/controllerserver.py", line 65, in CreateVolume
    volume = search_volume(request.name)
  File "/kadalu/volumeutils.py", line 734, in search_volume
    mount_glusterfs(volume, mntdir)
  File "/kadalu/volumeutils.py", line 928, in mount_glusterfs
    reload_glusterfs(volume)
  File "/kadalu/volumeutils.py", line 906, in reload_glusterfs
    execute(*cmd)
  File "/kadalu/kadalulib.py", line 119, in execute
    with subprocess.Popen(cmd,
  File "/usr/lib/python3.8/subprocess.py", line 854, in __init__
    self._execute_child(args, executable, preexec_fn, close_fds,
  File "/usr/lib/python3.8/subprocess.py", line 1637, in _execute_child
    self.pid = _posixsubprocess.fork_exec(
TypeError: expected str, bytes or os.PathLike object, not int
```

amarts:
Here is my take on this PR (as I am occupied for rest of the week).

Would like reviews on it, and more testing.

* Framework to enable remounting of all clients now added.
* TODO: add corresponding changes to kubectl-kadalu (CLI)
  - for both storage type changes and also decommissioned bricks.
* TODO: documentation
* TODO: testing (think how would it be better done).

@aravindavk @leelavg @vatsa287

leelavg:
> can't pass pid directly to cmd

Converting to string isn't working too?

> Would like reviews on it, and more testing

Occupied for this week, will try to squeeze sometime if possible.

csi/watch-vol-changes.sh
```diff
@@ -0,0 +1,17 @@
+#!/bin/bash
+
+CONFIG_MAP_INFO_DIR="/var/lib/gluster"
+
+ps aux | grep python
```
ook: Dead code.

csi/watch-vol-changes.sh
```diff
@@ -0,0 +1,17 @@
...
+
+CSI_PROCESS_ID=$(ps aux|grep python | grep main | awk '{ print $2}' | xargs);
+
+echo $CSI_PROCESS_ID;
+
+while true; do
```
ook, line 15: Should we trap sigup and kill that loop?

ook (commented):
Just comments, I didn't test that PR.

csi/watch-vol-changes.sh
amarts: Nope. The way I thought of it, without `-m` (or `--monitor`) option, the inotifywait breaks after 1st event. We expect this script to be running forever, and hence not planning to trap the signals here.

amarts: Added more comments to the file.

amarts:
Failed in same place, while doing 'ExpandVolume':

```
[2021-05-25 12:31:47,892] INFO [controllerserver - 265:CreateVolume] - Volume created	 name=sanity-expand-volume-3B5C57D3-D9AF33AF size=10737418240 hostvol=storage-pool-4 pvtype=subvol volpath=subvol/c9/91/sanity-expand-volume-3B5C57D3-D9AF33AF duration_seconds=5.09607195854187
[2021-05-25 12:31:47,935] DEBUG [volumeutils - 933:mount_glusterfs] - Already mounted	 mount=/mnt/storage-pool-1
[2021-05-25 12:31:47,944] DEBUG [volumeutils - 933:mount_glusterfs] - Already mounted	 mount=/mnt/storage-pool-4
[2021-05-25 12:31:47,949] INFO [controllerserver - 462:ControllerExpandVolume] - Existing PV size and Expansion requested PV size	 existing_pvsize=10737418240 expansion_requested_pvsize=11811160064
[2021-05-25 12:31:47,949] DEBUG [controllerserver - 475:ControllerExpandVolume] - Found PV type	 pvtype=subvol capability=
[2021-05-25 12:31:48,000] DEBUG [volumeutils - 427:is_hosting_volume_free] - pv stats	 hostvol=storage-pool-4 total_size_bytes=29249634304 used_size_bytes=21474836480 free_size_bytes=7774797824 number_of_pvs=3 required_size=11811160064 reserved_size=777479782.4
[2021-05-25 12:31:48,000] ERROR [controllerserver - 487:ControllerExpandVolume] - Hosting volume is full. Add more storage	 volume=storage-pool-4
```

It is picking disperse volume for ExpandVolume, is that an issue?

leelavg:
The error seems genuine, need to figure out below
- Can we limit expand volume size in csi-sanity?
- Is it intentional or randomly choosing a pvc to expand (need to look at actual code path of [csi-sanity](https://github.com/kubernetes-csi/csi-test/blob/v3.1.1/pkg/sanity/controller.go#L1905-L1952))

leelavg:
@amarts I wasn't able to reason through this [commit](https://github.com/kadalu/kadalu/pull/555/commits/bd672134be60bac3965d90699e06c35b149cf117) [wrt csi sanity] as it introduced new errors as per actions log.

Can you revert above and make below change if possible?

``` diff
diff --git a/tests/minikube.sh b/tests/minikube.sh
index fea9d04..e43ec8f 100755
--- a/tests/minikube.sh
+++ b/tests/minikube.sh
@@ -342,7 +342,8 @@ test_kadalu)
     kubectl wait --for=condition=ready pod -l app=sanity-app --timeout=15s
 
     exp_pass=33
-    kubectl exec sanity-app -i -- sh -c 'csi-sanity -ginkgo.v --csi.endpoint $CSI_ENDPOINT -ginkgo.skip pagination' | tee /tmp/sanity-result.txt
+    # Set expand vol size to 10MB
+    kubectl exec sanity-app -i -- sh -c 'csi-sanity -ginkgo.v --csi.endpoint $CSI_ENDPOINT -ginkgo.skip pagination -csi.testvolumesize 10485760' | tee /tmp/sanity-result.txt
 
     # Make sure no more failures than above stats
     act_pass=$(grep -Po '(\d+)(?= Passed)' /tmp/sanity-result.txt 2>/dev/null || echo 0)
```

csi/volumeutils.py
```diff
@@ -819,29 +823,48 @@ def generate_client_volfile(volname):
     with open(info_file_path) as info_file:
         data = json.load(info_file)
 
+    # If the hash of configmap is same for the given volume, then there
+    # is no need to generate client volfile again.
+    if not VOL_DATA.get(volname, None):
+        VOL_DATA[volname] = {}
```
amarts, line 829: This should get deleted in `delete_volume()`.

csi/volumeutils.py
```diff
@@ -914,7 +974,8 @@ def mount_glusterfs(volume, mountpoint, is_client=False):
             cmd.extend(["--client-pid", "-14"])
 
         try:
-            execute(*cmd)
+            (_, err, pid) = execute(*cmd)
+            VOL_DATA[volname]["pid"] = pid
```
amarts, line 988: Question @aravindavk : looks like the problem here is, pid is of the parent process (ie, glusterfs binary), but the process does fork() so this PID we received may not be running when we want to issue the SIGHUP.

In that case, I would do the kill sighup with a piped command to get the volume process based on name etc.. ?

aravindavk: thats good approach I think. Since the Storage names are unique, we can search the `ps` output to get the pid.

```
ps --no-header -ww -o pid,command -C glusterfs | grep "\-\-volfile-id=${volname}"
```

aravindavk:
```python
def get_volumes_mount_pids():
    cmd = ['ps', '--no-header', '-ww', '-o', 'pid,command', '-C', 'glusterfs']
    out, err = execute(*cmd)
    volumes = {}

    for line in out.split("\n"):
        parts = line.split()
        pid = parts[0]
        volname = ""
        for part in parts:
            if part.startswith("--volfile-id="):
                volname = part.split("=")[-1]

        volumes[volname] = pid

    return volumes
```

csi/volumeutils.py
```diff
@@ -855,6 +878,51 @@ def generate_client_volfile(volname):
...
+        try:
+            out, err, _ = execute(*cmd)
+            for line in out.split("\n"):
+                parts = line.split()
+                pid = parts[0]
+                for part in parts:
+                    if part.startswith("--volume-id="):
+                        if part.split("=")[-1] == volname:
+                            cmd1 = [ "kill", "-HUP", pid ]
+                            execute(*cmd1)
```
aravindavk, line 911: Move this to a function, so that early return(loop break) after sighup execution.

csi/volumeutils.py
```diff
@@ -866,6 +934,7 @@ def mount_glusterfs(volume, mountpoint, is_client=False):
 
     # Ignore if already glusterfs process running for that volume
     if is_gluster_mount_proc_running(volname, mountpoint):
+        reload_glusterfs(volume)
```
aravindavk, line 937: This call can raise exception, csi server will crash on one notification failure

csi/volumeutils.py
```diff
@@ -874,6 +943,7 @@ def mount_glusterfs(volume, mountpoint, is_client=False):
 
     # Ignore if already mounted
     if is_gluster_mount_proc_running(volname, mountpoint):
+        reload_glusterfs(volume)
```
aravindavk, line 946: same as above

csi/controllerserver.py
```diff
@@ -469,6 +469,20 @@ def ControllerExpandVolume(self, request, context):
...
 
+        if not request.capacity_range:
+            errmsg = "Capacity Range is empty and must be provided"
+            logging.error(errmsg)
+            context.set_details(errmsg)
+            context.set_code(grpc.StatusCode.INVALID_ARGUMENT)
+            return csi_pb2.ControllerExpandVolumeResponse()
+
+        if not request.capacity_range.required_bytes:
+            errmsg = "Capacity Range is empty and must be provided"
```
leelavg, line 480: Can this errmsg be tweaked not to sound same as above?

server/mount-glustervol
```diff
@@ -18,15 +18,33 @@ with open(info_file) as f:
```
leelavg, line 18: Question: Where is this file actually being used in the code, I missed this from the get go?
I can observe similar code in at-least two other places (in generating_client_volfile and in shd) and I believe handler methods and above might need some refactoring :/

templates/Replica1.client.vol.j2
```diff
@@ -10,6 +10,9 @@ end-volume
 volume {{ volname }}-dht
     type cluster/distribute
     subvolumes {{ dht_subvol|join(' ') }}
+{% if decommissioned != "" %}
+    option decommissioned-bricks {{ decommissioned }}
```
leelavg, line 14: If you don't mind, can you pls explain this?
- Like when a brick/subvol is decommissioned we'll be performing ~replace brick kind of operation and should wait for heal to complete right? (or)
- Is it just like we don't need (~data in) old bricks and they can be kinda replaced all together?

leelavg (commented):
Entirely optional, can this PR be broken into two as it was there from the start, one for volume change and another for decommission?

Rationale being, many regressions in CI and we aren't sure of the effects in untested code paths?

templates/Replica1.client.vol.j2
amarts: https://github.com/kadalu/glusterfs/blob/devel/xlators/cluster/dht/src/dht-shared.c#L960 

this just makes the brick not get any new data. But reads will be served from the same brick. There is no 'heal', but rebalance operation which is required.

server/mount-glustervol
amarts: Nope, this is not used anywhere IMO.

csi/controllerserver.py
amarts: Ack

csi/volumeutils.py
amarts: Done in #570

csi/controllerserver.py
amarts: Addressed in #571

Issue #562, opened 2021-05-24
[Tracker] Issues related to external volumes usage


leelavg:
Hi,

I think below issues are ~good to have fix and so creating this issue to track them.

- [x] Delete external storage classes gracefully
- [x] Make use of `gluster_options` while actually mounting gluster volume
- [ ] Minimal testing wrt external cluster, can gluster be dockerized and used in CI to iron out issues similar to #468 and #540?
- [ ] Revisit kadalu-quotad for any incomplete implementation wrt current code ([native gluster rfc](https://kadalu.io/rfcs/0007-Using-GlusterFS-directory-quota-for-external-gluster-volumes.html))
- [x] Revisit xattr's code that is being set on external vols (#526)
- [ ] In general, prioritise or at-least take note of any incompatibilities that may be introduced after adding features to kadalu storage

I'll add logs to all of these if possible at the earliest.

PR #505, opened 2021-03-30
Implementaion of ListVolumes in Controller Service

leelavg:
- Limit max_entries at 30 if CO didn't set any limit
- Support of paging and max_entries
- Tested locally
- Update sanity debug manifests

PR #564, opened 2021-05-27
Fix listvolumes and external volume deletions

### LINKED ISSUE #553: Minor issue seen with ListVolumes
If a storage config is added, but there is no PVC request made from that volume, then we get below failure.

```
[Fail] Controller Service [Controller Server] ListVolumes [It] should return appropriate values (no optional values added) 
/go/csi-test-3.1.1/pkg/sanity/controller.go:187

[Fail] Controller Service [Controller Server] ListVolumes [It] check the presence of new volumes and absence of deleted ones in the volume list 
/go/csi-test-3.1.1/pkg/sanity/controller.go:215
```

Logs from csi-provisioner

```
[2021-05-22 05:49:04,754] ERROR [_server - 445:_call_behavior] - Exception calling application: [Errno 2] No such file or directory: '/mnt/add-brick/info'
  File "/kadalu/lib/python3.8/site-packages/grpc/_server.py", line 435, in _call_behavior
  File "/kadalu/controllerserver.py", line 393, in ListVolumes
    next(GEN)
  File "/kadalu/volumeutils.py", line 1158, in yield_list_of_pvcs
    for idx, value in enumerate(wrap_pvc(yield_pvc_from_hostvol)):
  File "/kadalu/volumeutils.py", line 1145, in wrap_pvc
    prev = next(gen)
  File "/kadalu/volumeutils.py", line 1135, in yield_pvc_from_hostvol
    yield from pvcs
  File "/kadalu/volumeutils.py", line 1116, in yield_pvc_from_mntdir
    for child in os.listdir(mntdir):
FileNotFoundError: [Errno 2] No such file or directory: '/mnt/add-brick/info'
```

leelavg:
- Fix no pvc and no hostvol scenario in listvols
- Corner case still exist due to latency in mount/configmap/ENOTCON etc
however there's little to nill ROI fixing those at the moment
- Handle external volume(~only sc & cm) deletion and this should fix multiple provisioner
restarts while deleting external vol

Fixes: #553
Updates: #505, #562

### DIFF
--- a/csi/controllerserver.py
+++ b/csi/controllerserver.py
@@ -7,10 +7,9 @@ import os
 import random
 import time
 
-import grpc
-
 import csi_pb2
 import csi_pb2_grpc
+import grpc
 from kadalulib import logf, send_analytics_tracker
 from volumeutils import (HOSTVOL_MOUNTDIR, PV_TYPE_SUBVOL, PV_TYPE_VIRTBLOCK,
                          check_external_volume, create_subdir_volume,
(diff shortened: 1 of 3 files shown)


leelavg:
Testing Performed:
- Calling ListVolumes with & without PVC, with & without hostvol, with external vol & kadalu native vols
- Delete kadalustorage with external vol & kadalu native vols + with & without hosting PVC

Observation:
- RPC methods may fail when called just after modification to cm/mount/server-pod changes etc (kinda expected)
- Deletion of kadalustorage when pvc exists deletes resources created from kadalu crd however the volume isn't deleted
- When we try to delete it again kubectl will throw and error (workaround may not be possible) as our source of truth is ConfigMap for most if not all the operations.

There are some minor observations but most of them is due to unavoidable latency.

amarts (approved):

PR #517, opened 2021-04-17
Publish manifest files, helm chart and kubectl-kadalu files to gh-pages

aravindavk:
Once published to `gh-pages` branch of this repo, those files will be
available as `https://kadalu.io/kadalu/downloads/<file-name>`

For example, https://kadalu.io/kadalu/downloads/kubectl-kadalu

Signed-off-by: Aravinda Vishwanathapura <aravinda@kadalu.io>

PR #565, opened 2021-05-27
[no ci] fix actions-gh-pages

leelavg:
As per [this run](https://github.com/kadalu/kadalu/runs/2681773839?check_suite_focus=true#step:13:15), this action seems to be picking absolute path by default


leelavg:
It's failed again, and it seems the action is cloning [gh-pages](https://github.com/kadalu/kadalu/runs/2693348647?check_suite_focus=true#step:13:39) branch.

Original assumption (#517) might be incorrect, need to relook.

PR #572, opened 2021-06-01
allow modification of Replica1 type to Replica2 or Replica3

amarts:
Along with it, fix a bug which resulted in array index out of range.

### DIFF
--- a/operator/main.py
+++ b/operator/main.py
@@ -561,7 +561,7 @@ def handle_added(core_v1_client, obj):
 
     # Generate Node ID for each storage device.
     for idx, _ in enumerate(obj["spec"]["storage"]):
-        obj["spec"]["storage"][idx]["node_id"] = str(uuid.uuid1())
+        obj["spec"]["storage"][idx]["node_id"] = "node-%d" % idx
 
     update_config_map(core_v1_client, obj)
     deploy_server_pods(obj)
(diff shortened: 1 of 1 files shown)


operator/main.py
```diff
@@ -604,7 +595,8 @@ def handle_modified(core_v1_client, obj):
 
     # Set Node ID for each storage device from configmap
     for idx, _ in enumerate(obj["spec"]["storage"]):
-        obj["spec"]["storage"][idx]["node_id"] = cfgmap["bricks"][idx]["node_id"]
+        if not obj["spec"]["storage"][idx].get("node_id", None):
```
leelavg: Will it always be None? (or) something like `node_id = cfgmap["bricks"][idx].get("node_id") or str(uuid.uuid1())` would be better ig.

leelavg (commented):
Minor comment

operator/main.py
amarts: The bug fixed here is to not have `idx` on cfgmap['bricks'].

The usecase is this: I had a config with type ReplicaX/Disperse., with say M nodes. Now, I added X nodes, and did `kubeclt apply`. At this time, cfgmap entry would have say M entries, and obj would have M+X entries. Hence the out of range. 

But thinking more on this, everytime the config changes, the node_id changes. But considering it gets used in server config properly as server pods gets restarted, it is OK for us for now.

leelavg (approved):
LGTM

PR #580, opened 2021-06-21
Add link to latest artifacts

aravindavk:
Github provides URL alias to the latest releases.

```
https://github.com/[USER]/[REPO]/releases/latest/download/[FILENAME]
```

### DIFF
--- a/doc/quick-start.adoc
+++ b/doc/quick-start.adoc
@@ -6,7 +6,8 @@ Download the latest release of Kadalu Kubectl plugin using,
 
 [source,console]
 ----
-curl -LO https://github.com/kadalu/kadalu/releases/download/0.8.2/kubectl-kadalu
+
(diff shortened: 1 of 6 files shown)


doc/quick-start.adoc
```diff
@@ -6,7 +6,8 @@ Download the latest release of Kadalu Kubectl plugin using,
 
 [source,console]
 ----
-curl -LO https://github.com/kadalu/kadalu/releases/download/0.8.2/kubectl-kadalu
+
```
amarts, line 9: This is very important. Thanks for this.

While at it (or later), can we merge below 20+ lines and make a single block which can be copy pasted in one shot? This helps to make steps easier for people going through docs, mainly because the steps are self-explanatory, and the extra sentences are not adding any value.

amarts (approved):

leelavg (approved):
I think with this change we can remove stale gh-pages branch as well?

PR #581, opened 2021-06-21
remove dependency on pypi docker image

amarts:
* pypi images were added to save time to build on arm platform
  (specially for grpcio package). In latest versions, we see that
  it is taking just 2 mins, instead of 90mins. Hence its good to
  remove the dependency (and keep minimum variables)
* Also add a release tag to builder image, so we can debug it if
  needed.

### DIFF
--- a/.github/workflows/on-demand-pypi-source-push.yml
+++ b/.github/workflows/on-demand-pypi-source-push.yml
@@ -1,4 +1,4 @@
-name: Release
+name: PyPi images
 
 on:
   # Allow to run this workflow manually from the Actions tab
--- a/.github/workflows/on-release-tag.yml
+++ b/.github/workflows/on-release-tag.yml
@@ -1,4 +1,4 @@
-name: Release
+name: "On Release"
 
 on:
   push:
@@ -112,6 +112,7 @@ jobs:
           platforms: linux/amd64,linux/arm64,linux/arm/v7
           push: true
           tags: |
+            docker.io/kadalu/builder:${{ steps.vars.outputs.tag }}
             docker.io/kadalu/builder:latest
           build-args: |
             version=${{ steps.vars.outputs.tag }}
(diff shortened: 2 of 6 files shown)


aravindavk (approved):

amarts:
looks like `kubernetes 17.17` pypi didn't work... same issue as 12.x version. May be something we need to change? I will see if I can refresh the PR without that change, which should suffice.

.github/workflows/on-release-tag.yml
```diff
@@ -132,6 +133,7 @@ jobs:
             docker.io/kadalu/kadalu-csi:${{ steps.vars.outputs.tag }}
             docker.io/kadalu/kadalu-csi:latest
           build-args: |
+            builder_version=${{ steps.vars.outputs.tag }}
```
leelavg, line 136: Will the `builder_version` ever be different from `version`, if they are always same we can use only `version` here and in subsequent yml/Dockerfiles?

.github/workflows/on-demand-pypi-source-push.yml
```diff
@@ -1,4 +1,4 @@
-name: Release
+name: PyPi images
```
leelavg, line 1: Just confirming, this'll not be used anymore after this PR merge right?

leelavg (commented):
Just for my understanding, is there a chance that will we cross any limits in docker hub when we keep pushing builder image as well during every merge?

.github/workflows/on-demand-pypi-source-push.yml
amarts: Yes. But want to remove them in another PR.

amarts:
> Just for my understanding, is there a chance that will we cross any limits in docker hub when we keep pushing builder image as well during every merge?

We may. But for now, I am considering to get to that issue to resolve that issue. For now, not bother. We can also move to paid service to resolve issues for sometime. Lets face it when we get there.

.github/workflows/on-release-tag.yml
amarts: not yet sure. Lets keep it this way for a month or 2 and see where we can optimize, and how the other platform story evolves.

Issue #585, opened 2021-06-23
DNS and reserved ports issue on OpenShift 4.7


rmey:
I could not get kadalu.replica3 mode working with an OpenShift Cluster having 3 workers with attached storage on /dev/vdd each. In the logs I get DNS and reserved port warnings. The replication does not work - every of the 3 example pods which share a RWX PVC with replica3 mode storage class are looking like as they would have their own volume. The shared volume does not work.
Since the log is very big I created a gist:
[Full logs and example yamls](https://gist.github.com/rmey/6f162e2269c19f31fc2e096694aea51c)
```
...
[2021-06-23 09:04:29.086639 +0000] I [MSGID: 101190] [event-epoll.c:669:event_dispatch_epoll_worker] 0-epoll: Started thread with index [{index=0}] 
[2021-06-23 09:04:29.086726 +0000] I [MSGID: 101190] [event-epoll.c:669:event_dispatch_epoll_worker] 0-epoll: Started thread with index [{index=1}] 
[2021-06-23 09:04:32.062417 +0000] E [MSGID: 101075] [common-utils.c:519:gf_resolve_ip6] 0-resolver: error in getaddrinfo [{family=0}, {ret=Name or service not known}] 
[2021-06-23 09:04:32.062479 +0000] E [name.c:264:af_inet_client_get_remote_sockaddr] 0-storage-pool1-client-0: DNS resolution failed on host server-storage-pool1-0-0.storage-pool1
[2021-06-23 09:04:32.064563 +0000] E [name.c:264:af_inet_client_get_remote_sockaddr] 0-storage-pool1-client-1: DNS resolution failed on host server-storage-pool1-1-0.storage-pool1
[2021-06-23 09:04:32.067209 +0000] E [name.c:264:af_inet_client_get_remote_sockaddr] 0-storage-pool1-client-2: DNS resolution failed on host server-storage-pool1-2-0.storage-pool1
[2021-06-23 09:04:35.064844 +0000] E [name.c:264:af_inet_client_get_remote_sockaddr] 0-storage-pool1-client-0: DNS resolution failed on host server-storage-pool1-0-0.storage-pool1
[2021-06-23 09:04:35.066737 +0000] E [name.c:264:af_inet_client_get_remote_sockaddr] 0-storage-pool1-client-1: DNS resolution failed on host server-storage-pool1-1-0.storage-pool1
[...]

PR #588, opened 2021-06-28
Exception handling for ProtocolError in operator

leelavg:
Updates: #585

### DIFF
--- a/operator/main.py
+++ b/operator/main.py
@@ -15,6 +15,7 @@ from jinja2 import Template
 from kadalulib import execute as lib_execute
 from kadalulib import logf, logging_setup, send_analytics_tracker
 from kubernetes import client, config, watch
+from urllib3.exceptions import ProtocolError
 from utils import CommandError
 from utils import execute as utils_execute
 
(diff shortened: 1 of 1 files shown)


operator/main.py
```diff
@@ -838,28 +839,38 @@ def crd_watch(core_v1_client, k8s_client):
...
+                    handle_modified(core_v1_client, obj)
+                elif operation == "DELETED":
+                    handle_deleted(core_v1_client, obj)
+        except ProtocolError:
+            # It might so happen that this'll be logged for every hit in k8s
+            # stream and better to log at debug level
+            logging.debug(
+                logf(
+                    "Watch connection broken and restarting watch on the stream"
+                ))
```
aravindavk, line 880: Add little sleep(`time.sleep(1)`) here to avoid infinite loop consuming full cpu.

operator/main.py
```diff
@@ -838,28 +839,38 @@ def crd_watch(core_v1_client, k8s_client):
...
+                metadata = obj.get("metadata")
+                resource_version = metadata['resourceVersion']
+                logging.debug(
+                    logf("Event", operation=operation, object=repr(obj)))
+                if operation == "ADDED":
+                    handle_added(core_v1_client, obj)
+                elif operation == "MODIFIED":
+                    handle_modified(core_v1_client, obj)
+                elif operation == "DELETED":
+                    handle_deleted(core_v1_client, obj)
```
aravindavk: May be good to make this a function?

operator/main.py
leelavg: sure, do you think adding a threshold counter for number of tries would also be good? Like raise exception after ~10 times hitting the exception and the operator will be restarted?

operator/main.py
leelavg: ack.

operator/main.py
aravindavk: Restarting may not help if the problem persists. Increase the sleep duration to 30 seconds?

leelavg: ack

amarts (approved):

aravindavk (approved):

Issue #587, opened 2021-06-27
[RFE] Heketi to Kadalu Migration


aravindavk:
Recently Volume ID option is added while adding Kadalu Storage. This is good enough to migrate the Heketi Volumes to Kadalu Storage. But once imported it will be used as common storage pool and new PVs may gets assigned from this Storage.

Proposal:

* Get the Gluster Volume ID, PV ID(Ex: `pvc-ed2e3167-5040-4ab0-9b6c-a2dbcc552b6e`), Volume Type and bricks list
* Join the Heketi nodes to Kubernetes Cluster and note the hostnames by running `kubectl get nodes` command.
* Note the device paths(`LVs`)
* Create Kadalu Storage(Using `storage-add`) with the name same as PV ID. Specify the Storage disks in the same order of bricks in Heketi managed Volume. **Note**: Use `--volume-id` option to specify the Heketi managed Gluster Volume ID.
* Change CSI Provisioner to not use Kadalu Storages if the name starts with `pvc-`
* Change CSI Provisioner to map the Kadalu Storage if PV ID matches the Kadalu Storage name(No Quota, Use full Volume)
* Create PV with the same name as used with Heketi.
* Change CSI node plugin to mount the Kadalu Storage root directory when a Pod requests the mount of a PV with ID same as Kadalu Storage name.

Many things can be automated, for example: Use the `kubectl` and Heketi CLI to get the Volume info and convert the Heketi Volume info to `kubectl kadalu` command.

amarts:
I suggest, treating this and External storage with non-native format in a similar way. Way less code, and similar handling for heketi volume.

leelavg:
My knowledge about Heketi is rather limited however I'm unsure whether block hosting volumes can also be migrated in the same or not 🤔

@amarts I guess we can map heketi independent mode to external non-native and everything else to kadalu native as they are deployed as pods already or count everything as External for a start?

aravindavk:
> I suggest, treating this and External storage with non-native format in a similar way. Way less code, and similar handling for heketi volume.

Makes sense. But when all the Storage nodes are migrated to Kadalu then it should be handled as native way with a few changes to Provisioner code. But node plugin can share the code of non-native format.

leelavg:
@aravindavk small query, did you design a way to move all volume handling logic to operator and serve through REST API (first referenced in https://github.com/kadalu/kadalu/issues/553#issuecomment-848597576)?

amarts:
[...]

PR #589, opened 2021-06-28
operator: Add support for deployment of custom storageclass

vatsa287:
This PR adds support for deployment of custom storage classes with name associated with storage-pool-name, and pre-selects the storage for PVC Binding with the same storage-pool-name.


operator/main.py
```diff
@@ -211,10 +212,13 @@ def validate_volume_request(obj):
                       " specified")
         return False
 
+    # should'nt this come under disperse volume?? Not replica??
```
aravindavk: Are these questions to reviewers? :P

vatsa287: Sorry I had to remove this before updating here !!

operator/main.py
```diff
@@ -211,10 +212,13 @@ def validate_volume_request(obj):
...
     if subvol_bricks_count > 1:
+        # Why not range(0, 1): since / is always == 1 ??
```
aravindavk: More than 1 when the Storage is "Distributed Replicate" or "Distributed Disperse"

operator/main.py
```diff
@@ -211,10 +212,13 @@ def validate_volume_request(obj):
...
         for i in range(0, int(len(bricks) / subvol_bricks_count)):
             decommissioned = ""
             for k in range(0, subvol_bricks_count):
+                # why not brick_idx = k since value is always same, and i always 0
```
aravindavk: Example: Replica count = 3, Number of bricks = 6. Inner loop iterates 0, 1, 2 and 0, 1, 2 two times.

with below line the index will be 0, 1, 2, 3, 4, 5

operator/main.py
vatsa287: Thanks

operator/main.py
vatsa287: Thanks alot got it

vatsa287:
### Test Run
Adding 5 storages, deploys 5 custom scs
```
sudo kubectl get sc -nkadalu
NAME                 PROVISIONER                RECLAIMPOLICY   VOLUMEBINDINGMODE   ALLOWVOLUMEEXPANSION   AGE
kadalu               kadalu                     Delete          Immediate           true                   29m
kadalu.disperse      kadalu                     Delete          Immediate           true                   29m
kadalu.replica1      kadalu                     Delete          Immediate           true                   29m
kadalu.replica2      kadalu                     Delete          Immediate           true                   29m
kadalu.replica3      kadalu                     Delete          Immediate           true                   29m
kadalu.sp-1          kadalu                     Delete          Immediate           true                   29m
kadalu.sp-2          kadalu                     Delete          Immediate           true                   27m
kadalu.sp-3          kadalu                     Delete          Immediate           true                   27m
kadalu.sp-4          kadalu                     Delete          Immediate           true                   27m
kadalu.sp-5          kadalu                     Delete          Immediate           true                   27m
standard (default)   k8s.io/minikube-hostpath   Delete          Immediate           false                  107d
```
Adding a PVC for sp-3 gets binded to it
```
[2021-06-28 05:39:28,941] INFO [controllerserver - 300:CreateVolume] - Volume created	 name=pvc-2651bb0a-9d05-4406-8e49-4bb65ce461a8 size=104857600 hostvol=sp-3 pvtype=subvol volpath=subvol/32/36/pvc-2651bb0a-9d05-4406-8e49-4bb65ce461a8 duration_seconds=0.4838991165161133
``` 
The PVC Yaml file
```
# -*- mode: yaml -*-
---
kind: PersistentVolumeClaim
apiVersion: v1
metadata:
  name: pv3
spec:
  storageClassName: kadalu.sp-3
  accessModes:
    - ReadWriteOnce # You can also provide 'ReadWriteMany' here
  resources:
    requests:
      storage: 100Mi
```

leelavg:
hi @vatsa287 

Pardon, I may be missing the point here and so can you please address below if possible?

- We already have multiple storage classes deployed and adding one more option may confuse users and I hope this PR doesn't call for any cli changes?
- Why are we deploying storage classes without validating it first? (`deploy_storage_class` is moved to start of the function)
- We are already documenting process for creating custom storage classes in https://kadalu.io/docs/k8s-storage/latest/storage-classes/ and how different is PR to that?

In my opinion it's better to reduce the storage classes and only have single kadalu provisioner 😅 

@aravindavk / @amarts I'm just stating my opinions and not against any functionality improvements if you feel otherwise.

amarts:
>  @amarts I'm just stating my opinions and not against any functionality improvements if you feel otherwise.

Good questions. I will clarify some of this as I am the one who suggested @vatsa287 to work on this PR.

* About existing 'multiple' storage classes.

The current storage classes like replica{1,2,3} were designed when kadalu didn't had support for distribute translator of glusterfs in its deployment, and we needed a mechanism to use more than one storage in a deployment. Grouping the storage configs under Replica1,2,3 allowed us to tell user to use one of these to use PV from. While that is 'still' a valid usecase, with support of distribute the chances of adding multiple 'similar' storage nodes as separate volume is _almost_ non-existent, as they can all use  a single gluster volume to utilize all the nodes/storages. Again, this won't go to very high number now, as unlike heketi usecases, we won't be consuming 1 glusterfs volume for 1 PV.

* About having document

What I felt is, even though we 'document' about how to consume a particular volume of glusterfs as a storage class, it is not directly available, and may not be intuitive for many users. My take is, gluster's users are used to relating all storage operations based on 'volume' (which is similar in other storage products too). So, even here, the naming convention to use same storage volume in a given PV would be adhered to, if we make it automatic, instead of allowing users to mess around with names.

Let me know if these give some idea about what we are trying to do. This along with #587 will make PVC to use the relavant storage classes to get into proper glusterfs volume.

leelavg:
@amarts thanks for taking time to fill the gaps in my understanding, yes this certainly helped.

vatsa287:
@leelavg Moving the deployment of StorageClass after volume validation

leelavg (approved):
Minor optimisations are possible however as we are expecting more changes, this is good to go. LGTM.

operator/scutils.py
```diff
@@ -0,0 +1,18 @@
+from string import Template
```
aravindavk: For uniformity, use the Jinja2 template itself similar to other templates

operator/scutils.py
```diff
@@ -0,0 +1,18 @@
...
+
+CUSTOM_SC_TEMPLATE = """kind: StorageClass
```
aravindavk: Move this to templates dir `storageclass.yaml` or some other name

operator/scutils.py
vatsa287: Done

operator/scutils.py
vatsa287: I have used storageclass-kadalu.custom.yaml and it is executed along with default storage classes. The sc is deleted when hosting volume is deleted. The older data of hosting volume is overwritten and deployed for next hostvol.

templates/storageclass-kadalu.custom.yaml.j2
```diff
@@ -0,0 +1,11 @@
+# -*- mode: yaml -*-
+---
+kind: StorageClass
+apiVersion: storage.k8s.io/v1
+metadata:
+  name: kadalu.{{ hostvol_name }}
+provisioner: kadalu
+allowVolumeExpansion: true
+parameters:
+  hostvol_type: {{ type }}
```
aravindavk: This can be removed

operator/main.py
```diff
@@ -974,8 +997,8 @@ def deploy_storage_class():
             logging.info(logf("StorageClass already present, continuing with Apply",
                               manifest=filename))
 
-        # Deploy Storage Class
-        template(filename, namespace=NAMESPACE, kadalu_version=VERSION)
+        template(filename, namespace=NAMESPACE, kadalu_version=VERSION,
```
aravindavk, line 997: Remove all other storage class files except the last one

```
storageclass-kadalu.disperse.yaml.j2
storageclass-kadalu.replica1.yaml.j2
storageclass-kadalu.replica2.yaml.j2
storageclass-kadalu.replica3.yaml.j2

storageclass-kadalu.yaml.j2
```

amarts: Wont we cause regression for existing users if we remove it? Or because they have already set it up like this, no need for these in future versions?

aravindavk: Already deployed SCs will not be touched, no new SCs for these types are created.

templates/storageclass-kadalu.custom.yaml.j2
vatsa287: Removed hostvol_type. When the type is removed, how to specify repllica1 or any other required type?

operator/main.py
```diff
@@ -662,18 +665,16 @@ def handle_deleted(core_v1_client, obj):
...
-            lib_execute(KUBECTL_CMD, DELETE_CMD, "sc", volname)
-            logging.info(logf(
-                "Deleted Storage class",
-                volname=volname,
-            ))
+            delete_storage_class(volname, hostvol_type)
             delete_config_map(core_v1_client, obj)
 
         else:
+            delete_storage_class(volname, hostvol_type)
```
amarts: nitpick: For simplicity, this line and `delete_config_map()` line can move out of the if/else logic as its same code in here.

amarts (approved):
This  PR makes many part of documentation obsolete. We need a quick PR to fix docs to remove `replica1` and similar `storageclass`  definition there.

operator/main.py
vatsa287: Done

amarts (approved):

leelavg (approved):
LGTM, only ask is can you construct any manifest using kadalu-cli and deploy (very basic sanity check, that'd be enough).

aravindavk (approved):

PR #590, opened 2021-06-30
Handle corner case only when storage is created but no PV

leelavg:
- This corner case is inhibiting deletion of server pods when no PV is created
- May not happen unless a user creates a wrong storage pool and tries to delete the pool without creating any PV
- In that scenario `pv_stats` table isn't created and current logic prohibits deletion of server pods

### DIFF
--- a/operator/main.py
+++ b/operator/main.py
@@ -745,6 +745,11 @@ def get_num_pvs(storage_info_data):
         return pv_count
 
     except CommandError as msg:
+        # If storage is created but no PV is carved then pv_stats table is not
+        # created in SQLITE3
+        if 'no such table' in msg:
+            # We are good to delete server pods
+            return 0
         logging.error(
             logf("Failed to get size details of the "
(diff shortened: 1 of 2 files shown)


leelavg:
- We are using `KeyboardInterrupt` to stop the server and better to skip the pylint check for that
- I upgraded local pylint to match github runner but pylint checks are still failing 😞

operator/main.py
```diff
@@ -745,6 +745,11 @@ def get_num_pvs(storage_info_data):
         return pv_count
 
     except CommandError as msg:
+        # If storage is created but no PV is carved then pv_stats table is not
+        # created in SQLITE3
+        if msg.stderr.find("no such table") != 1:
```
aravindavk: `!= -1`

leelavg: ack, forgot to update important piece while looking into pylint. my bad 😥

aravindavk (approved):

PR #592, opened 2021-07-01
Pin Python Packages via requirements file for each component

### LINKED ISSUE #528: Pinning python packages to check against Pylint etc
Rationale:
- We are using all python modules inside containers and we have already pinned base container versions
- Whenever github runner updates it'll install all latest available packages as per requirements.txt which fails pylint
- We don't want to stop consuming latest github runner however it's better to pin versions in requirements.txt to almost resemble what's running in container
- We can also pin versions in Dockerfiles as well and raise separate PR for updating dependencies

leelavg:
Fixes: #528


leelavg:
1. Thought it'd be better to copy some rather than using same requirements file (like operator in ci_merge etc)
2. If we want to upgrade dependencies we can just run ~the command stated in the requirements file/setup.cfg~ `make gen-requirements` from root of the repo
3. Move pyxattr to builder?

amarts (commented):
PR looks good. Only question is, should we add all these files at the top level or is there a possibility to add them from a 'extras/' dir? Asking, so README in the repo is visible with lesser crawl :-)

No technical issues with the PR.

leelavg:
- Initial commit had files in different directories but they are failing due to docker directory context 
- Will try to move the files to extras/requirements folder (create) and send that to docker context in build command

leelavg:
@amarts 
> possibility to add them from a 'extras/' dir

created a new dir 'requirements' and added ~generated txt~ all files there as 'extras/' dir isn't considered in pr workflows

amarts (approved):

PR #597, opened 2021-07-09
CSI: allow kadalu_format 'native' or 'non-native' in all config

amarts:
* also remove '.external' part from storageclass name for external
  as it doesn't make sense anymore after merging #589

### DIFF
--- a/doc/external-gluster-storage.adoc
+++ b/doc/external-gluster-storage.adoc
@@ -7,7 +7,7 @@ server processes are not managed by kadalu operator.
 
 In both the ways please note `metadata.name` (let's say `e-vol`) in
 `kind: KadaluStorage` will create a StorageClass with name
-`kadalu.external.<volname>` (`volname` will be replaces with `e-vol`).
+`kadalu.<volname>` (`volname` will be replaces with `e-vol`).
 
 == Using external gluster in kadalu native way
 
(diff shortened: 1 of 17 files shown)


amarts:
@vatsa287 with this, if you handle `kadalu_format` check in CreateVolume in even non `External` type volumes, we can resolve #587 much smoothly.

@aravindavk what do you think.

leelavg (commented):
- Removing 'external' keyword is a good idea
- IMHO, it's better to have kadalu_format only for external storage as kadalu storages by default use subdir
- Even if we consider Heketi, I think those will fall under external non-native 🤔 

Nevertheless, if we still want to have kadalu_format better we have only one option at `spec` level in CRD and remove it from `details`. In addition to that, below things should be taken care
1. https://github.com/kadalu/kadalu/blob/devel/cli/kubectl_kadalu/storage_add.py#L121
2. https://github.com/kadalu/kadalu/blob/devel/cli/kubectl_kadalu/storage_add.py#L277
3. https://github.com/kadalu/kadalu/blob/devel/operator/main.py#L492

amarts:
> Even if we consider Heketi, I think those will fall under external non-native thinking 

It makes sense to provide kadalu native way for migration, that way, it would become users' choice to use it either way. (We already have option of external 1-1 PV).

aravindavk (approved):
LGTM

vatsa287 (approved):

doc/external-gluster-storage.adoc
```diff
@@ -31,14 +31,14 @@ metadata:
   name: ext-config
 spec:
   type: External
+  # Omitting 'kadalu_format' or using 'native' as a value will create
```
vatsa287, line 34: Will not kadalu_format: non-native create 1PV = 1Subdir? So shouldn't value above should be non-native?

PR #602, opened 2021-07-21
handle kadalu_format 'non-native' in kadalu volumes also

amarts:

### DIFF
--- a/csi/controllerserver.py
+++ b/csi/controllerserver.py
@@ -137,6 +137,7 @@ class ControllerServer(csi_pb2_grpc.ControllerServicer):
 
             hostvoltype = data['type']
 
+        kformat = filters.get('kadalu_format', "native")
         if hostvoltype == 'External':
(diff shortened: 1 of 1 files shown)


csi/controllerserver.py
```diff
@@ -289,6 +289,24 @@ def CreateVolume(self, request, context):
...
+            return csi_pb2.CreateVolumeResponse(
+                volume={
+                    "volume_id": request.name,
+                    "capacity_bytes": pvsize,
+                    "volume_context": {
+                        "type": hostvoltype,
+                        "hostvol": hostvol,
+                        "pvtype": pvtype,
+                        "fstype": "xfs",
+                        "kadalu_format": "non-native"
```
leelavg: I think it's a good idea to use this field (`kadalu_format`) in other responses (3 other places in same fn) as well, just to be consistent?

amarts: Done

PR #603, opened 2021-07-26
doc: add details of disabling simple-quota

amarts:

### DIFF
--- a/doc/troubleshooting.adoc
+++ b/doc/troubleshooting.adoc
@@ -95,3 +95,19 @@ If you have installed `kubectl_kadalu` package, then you can do below to get the
 ----
 $ kubectl kadalu logs
 ----
+
+
+=== Quota of PVCs
+
+`kadalu` uses simple-quota feature of glusterfs, which is present only in kadalu storage releases of glusterfs.
+
+As this is a new feature of glusterfs, there is possibilities where an user can hit a bug which is stopping the usage in in production. Hence, we have provided an option to disable quota limit check on PVCs of a particular storage pool. Please use below steps to get this working.
+
+
+[source,console]
+----
+$ kubectl exec -it kadalu-csi-provisioner-0 -c kadalu-provisioner -- bash
+# setfattr -n glusterfs.quota.disable-check -v "1" /mnt/${storage-pool-name}
+----
+
+This disable check is 'runtime' only fix right now, so if the server pods are restarted, this command may need to be issued again. Similarly to enable the check again, just pass the value as `"0"`.
--- a/server/Dockerfile
+++ b/server/Dockerfile
@@ -19,7 +19,7 @@ COPY --from=builder /opt /opt
 COPY --from=builder /kadalu /kadalu
 
 RUN apt-get update -yq && \
-    apt-get install -y --no-install-recommends python3 xfsprogs libtirpc3 sqlite3 && \
+    apt-get install -y --no-install-recommends attr python3 xfsprogs libtirpc3 sqlite3 && \
     apt-get -y clean && \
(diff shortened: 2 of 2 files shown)


doc/troubleshooting.adoc
```diff
@@ -95,3 +95,19 @@ If you have installed `kubectl_kadalu` package, then you can do below to get the
...
+
+`kadalu` uses simple-quota feature of glusterfs, which is present only in kadalu storage releases of glusterfs.
+
+As this is a new feature of glusterfs, there is possibilities where an user can hit a bug which is stopping the usage in in production. Hence, we have provided an option to disable quota limit check on PVCs of a particular storage pool. Please use below steps to get this working.
+
+
+[source,console]
+----
+$ kubectl exec -it kadalu-csi-provisioner-0 -c kadalu-provisioner -- bash
+# setfattr -n glusterfs.quota.disable-check -v "1" /mnt/${storage-pool-name}
```
aravindavk, line 110: `#` is extra in the line beginning?

amarts: When you call bash, it gives a shell with '#' and hence I added it.

doc/troubleshooting.adoc
```diff
@@ -95,3 +95,19 @@ If you have installed `kubectl_kadalu` package, then you can do below to get the
 ----
 $ kubectl kadalu logs
 ----
+
+
+=== Quota of PVCs
+
+`kadalu` uses simple-quota feature of glusterfs, which is present only in kadalu storage releases of glusterfs.
+
+As this is a new feature of glusterfs, there is possibilities where an user can hit a bug which is stopping the usage in in production. Hence, we have provided an option to disable quota limit check on PVCs of a particular storage pool. Please use below steps to get this working.
```
leelavg, line 104: May be a dumb question, does it effectively mean 1 vol : 1 pv?

doc/troubleshooting.adoc
aravindavk: my bad. got it. I looked at it as a single command to execute from outside the pod itself

```
kubectl exec -it kadalu-csi-provisioner-0 -c kadalu-provisioner -- setfattr -n glusterfs.quota.disable-check -v "1" /mnt/${storage-pool-name}
```

doc/troubleshooting.adoc
aravindavk: No. Multiple PVs, but PVs can overuse the storage than they are provisioned. This is only a workaround to unblock the Production issue.

leelavg: Ack, in a way I was also getting to that point.

I feel there should be a note that this section will be revised by next release (or at the earliest possible). 

I can already think of cases where it'll create confusion to users and debugging burden on us.

Please don't mind, if it hadn't been a prod issue, I would've strictly nack'd it. I guess you also feel the same😅

leelavg (approved):
LGTM

server/Dockerfile
```diff
@@ -19,7 +19,7 @@ COPY --from=builder /opt /opt
 COPY --from=builder /kadalu /kadalu
 
 RUN apt-get update -yq && \
-    apt-get install -y --no-install-recommends python3 xfsprogs libtirpc3 sqlite3 && \
+    apt-get install -y --no-install-recommends attr python3 xfsprogs libtirpc3 sqlite3 && \
```
aravindavk, line 22: nit: is this new dependency used in any code?

leelavg: I do not see `getfattr` in `server` pod so Amar is adding this.

We can get away with python xattr but i feel, this is a good addition.

aravindavk (approved):

Issue #577, opened 2021-06-15
Disk usage miscalculated / quota exceeded


benedekl:
Dear guys,

First, thank you for this awesome storage solution.

I have made a 3 node (Replica3) setup on Kubernetes. The bricks are on the master nodes. I'm using `docker.io/kadalu/kadalu-operator:devel` image as operator and tweaked the servers to tolerate `node-role.kubernetes.io/master` taints.

```yaml
apiVersion: kadalu-operator.storage/v1alpha1
kind: KadaluStorage
metadata:
  name: storage-pool-1
  namespace: kadalu
spec:
  storage:
    - device: /dev/vg0/gluster
      node: master-00
    - device: /dev/vg0/gluster
      node: master-01
    - device: /dev/vg0/gluster
      node: master-02
  type: Replica3
```

```
$ kubectl get nodes
NAME        STATUS   ROLES                  AGE   VERSION
master-00   Ready    control-plane,master   74d   v1.21.0
master-01   Ready    control-plane,master   74d   v1.21.0
master-02   Ready    control-plane,master   74d   v1.21.0
worker-00   Ready    <none>                 17d   v1.21.0
```

```
$ kubectl get pods -n kadalu -o wide
NAME                          READY   STATUS    RESTARTS   AGE   IP             NODE        NOMINATED NODE   READINESS GATES
kadalu-csi-nodeplugin-tbg6v   3/3     Running   0          25h   10.112.4.186   worker-00   <none>           <none>
kadalu-csi-provisioner-0      5/5     Running   0          25h   10.112.4.187   worker-00   <none>           <none>
operator-cb9d78479-gc9fj      1/1     Running   0          25h   10.112.4.189   worker-00   <none>           <none>
server-storage-pool-1-0-0     1/1     Running   0          25h   10.112.0.74    master-00   <none>           <none>
server-storage-pool-1-1-0     1/1     Running   0          25h   10.112.1.46    master-01   <none>           <none>
server-storage-pool-1-2-0     1/1     Running   0          25h   10.112.2.45    master-02   <none>           <none>
```

I tried to create a simple webserver deployment with PVC and everything seems fine. The PV has the correct size. I can exec into the pod and can create a 200M file on a 200M PV. After removing the file the used space goes back to 0.

[...]

PR #605, opened 2021-07-28
on-release: push operator image as last one

amarts:
Currently the release script is taking ~2hrs to complete, which
means the CLI would be ready to download before the images. As
our upgrade procedure is all about restarting the operator image
with the latest version, the operator image should be the last
one to be pushed to repository, to prevent any attempt of upgrades
within that time. This also prevents any 'data' layer access issues
during upgrade.

Updates: https://github.com/kadalu/kadalu/issues/577#issuecomment-887730673

### DIFF
--- a/.github/workflows/on-release-tag.yml
+++ b/.github/workflows/on-release-tag.yml
@@ -139,25 +139,6 @@ jobs:
           secrets: |
             KADALU_VERSION=${{ steps.vars.outputs.tag }}
       -
-        name: Build Operator Image and push
-        uses: docker/build-push-action@v2
-        timeout-minutes: 120
-        with:
-          context: .
-          file: operator/Dockerfile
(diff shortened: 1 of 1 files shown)


leelavg:
How'd it be to move `multi-arch-build-for-release` all the way up? Then we'll be releasing assets only after pushing images.

amarts:
> How'd it be to move `multi-arch-build-for-release` all the way up? Then we'll be releasing assets only after pushing images.

https://github.com/kadalu/kadalu/actions/runs/1071870525 (they are executed in parallel). So, this is good enough now.

Other fix is, instead of making it a separate job, make all as part of same job.

amarts:
@KyleSanderson this PR should address the concern you raised (which ofcourse is very valid).

aravindavk (approved):

Issue #596, opened 2021-07-09
RFE: Allow kadalu to manage external volume quota ( gluster-native directory)


hunter86bg:
It seems that kadalu can assign quota to PVCs initially on external volumes (native gluster directory quota), but has no option to extend it via kubectl.

Can we enable kadalu to change gluster's dir quota upon resize.
The feature from Kubernetes side is: https://kubernetes.io/blog/2018/07/12/resizing-persistent-volumes-using-kubernetes/

amarts:
Code to make changes (if someone is interested to pick this up).

in `update_subdir_volume()` (csi/volumeutils.py), handle the case separately for External volume (refer CreateVolume code in csi/controllerserver.py), and send a quota limit set command to cluster (with changed quota)).

hunter86bg:
@amarts , I can't find csi/controller.py . Maybe you ment csi/controllerserver.py ?

amarts:
> @amarts , I can't find csi/controller.py . Maybe you ment csi/controllerserver.py ?

@hunter86bg sorry about that. Yes. You are right. Editing my detail above.

hunter86bg:
@amarts , @aravindavk , maybe this one can be closed ?

leelavg:
RFE is implemented and verified via #607 and #619

PR #607, opened 2021-07-28
csi: provide proper Expansion option to External volume

amarts:
Updates: #596 

### DIFF
--- a/csi/controllerserver.py
+++ b/csi/controllerserver.py
@@ -328,7 +328,7 @@ class ControllerServer(csi_pb2_grpc.ControllerServicer):
             duration_seconds=time.time() - start_time
         ))
 
-        update_free_size(hostvol, hostvoltype, request.name, -pvsize)
+        update_free_size(hostvol, request.name, -pvsize)
 
         send_analytics_tracker("pvc-%s" % hostvoltype, uid)
         return csi_pb2.CreateVolumeResponse(
@@ -545,7 +545,7 @@ class ControllerServer(csi_pb2_grpc.ControllerServicer):
 
         # Get existing volume
         existing_volume = search_volume(request.volume_id)
-        if existing_volume.kformat == 'non-native':
+        if existing_volume.extra['kformat'] == 'non-native':
             errmsg = "PV with kadalu_format == non-native doesn't support Expansion"
             logging.error(errmsg)
             # But lets not fail the call, and continue here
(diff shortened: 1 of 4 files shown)


amarts:
Haven't yet tested. Need some help in testing this.

csi/controllerserver.py
```diff
@@ -590,7 +596,42 @@ def ControllerExpandVolume(self, request, context):
...
+                "gluster",
+                "volume",
+                "quota",
+                "%s" % existing_volume.gvolname,
+                "limit-usage",
+                "/%s" % existing_volume.volpath,
+                "%s" % expansion_requested_pvsize * 0.05,
+            ]
+            try:
+                execute(*quota_cmd)
```
aravindavk: Repeat code can be moved to a function

csi/controllerserver.py
```diff
@@ -590,7 +596,42 @@ def ControllerExpandVolume(self, request, context):
...
+            except CommandException as err:
+                errmsg = "Unable to set Gluster Quota via ssh"
+                logging.error(logf(errmsg, error=err))
+                context.set_details(errmsg)
+                context.set_code(grpc.StatusCode.INVALID_ARGUMENT)
+                return csi_pb2.CreateVolumeResponse()
```
aravindavk: return csi_pb2.ControllerExpandVolumeResponse()

csi/controllerserver.py
```diff
@@ -545,6 +545,11 @@ def ControllerExpandVolume(self, request, context):
 
         # Get existing volume
         existing_volume = search_volume(request.volume_id)
+        if existing_volume.extra['kformat'] == 'non-native':
```
leelavg, line 562: How about setting `allowVolumeExpansion` dynamically [here](https://github.com/kadalu/kadalu/blob/devel/templates/storageclass-kadalu.custom.yaml.j2#L8) and [here](https://github.com/kadalu/kadalu/blob/devel/templates/external-storageclass.yaml.j2#L7) while deploying `sc` itself?

Then we might get a good side effect of getting this field reflected in `kubectl get pvc` and k8s will not allow expansion as well. Feasible?

csi/controllerserver.py
leelavg: ig, it's better to have an `else` as well, like some users may not use gluster quota and in that case exapandvol should just log a msg?

aravindavk (approved):

csi/controllerserver.py
amarts: allowVolumeExpansion is not present? regardless, should be a separate PR IMO.

amarts: After checking, yes, it should be added with this PR itself. Done now. Thanks.

amarts:
@yugyama if you get chance, can you also have a look on this PR? Any concerns?

yugyama:
> @yugyama if you get chance, can you also have a look on this PR? Any concerns?

Looks good to me

Issue #606, opened 2021-07-28
All are PVC disconnected from pods during upgrade


couloum:
Hello,

We have performed the upgrade from `v0.8.3` to `v0.8.4` today in our preview environment. During the upgrade, all pods with a PVC attached got their PVC "disconnected". When we tried to access filesystem from the pod, we got this error: 

```
root@mysql-6f4d9fdcc5-8brk4:/# df -h
df: /var/lib/mysql: Transport endpoint is not connected
```

We had to delete all pods. After recreation, the PVC is correctly attached to new pods.

Is this the attended behavior?

Thank you

couloum:
Hi,

As an additional note, I have observed that PVC get disconnected when the `csi-nodeplugin` instance running on the same node as the pod is being restarted.

Here are log of `csi-nodeplugin` on the same node as a pod that got it's transport endpoint disconnected:

```
# ktailf --tail 50 kadalu-csi-nodeplugin-mh8fg csi-node-driver-registrar
I0728 19:32:38.614787       1 main.go:112] Version: v2.0.1-0-g1aaa13a-dirty
I0728 19:32:38.614866       1 main.go:122] Attempting to open a gRPC connection with: "/plugin/csi.sock"
I0728 19:32:38.618185       1 connection.go:151] Connecting to unix:///plugin/csi.sock
I0728 19:32:39.637624       1 main.go:129] Calling CSI driver to discover driver name
I0728 19:32:39.637674       1 connection.go:180] GRPC call: /csi.v1.Identity/GetPluginInfo
I0728 19:32:39.637689       1 connection.go:181] GRPC request: {}
I0728 19:32:39.644735       1 connection.go:183] GRPC response: {"name":"kadalu","vendor_version":"0.8.4"}
I0728 19:32:39.644840       1 connection.go:184] GRPC error: <nil>
I0728 19:32:39.644854       1 main.go:139] CSI driver name: "kadalu"
I0728 19:32:39.644890       1 node_register.go:55] Starting Registration Server at: /registration/kadalu-reg.sock
I0728 19:32:39.645129       1 node_register.go:64] Registration Server started at: /registration/kadalu-reg.sock
I0728 19:32:39.645245       1 node_register.go:86] Skipping healthz server because port set to: 0
I0728 19:32:40.135968       1 main.go:79] Received GetInfo call: &InfoRequest{}
[...]

PR #608, opened 2021-07-29
handle csi-nodeplugin install and management as a separate step

amarts:
If nodeplugins are restarted upon operator upgrade, there is a
possibility that PVs would be not able to function with FUSE
based filesystems. The process of upgrade (ie, to new version)
would be to update the operator by updating the operator with
newer version, and for client (ie, application/mount point)
upgrades, ideal process is to plan it as combined steps of node
drain, csi-node plugin upgrade, and then mark the node to pick
the applications. This way, there would be no application
dataloss.

Updates: #606


amarts:
@Madhu-1 - tagging you here as a reference check on this PR. Have a look when you can spare some time. I referenced your ceph-csi issue for this. Thanks.

Madhu-1:
Thanks, @amarts, sure I will take a look at it soon!

amarts:
note that 'helm' will be broken with this. As they would need to do the step of installing csi-nodeplugin with `kubectl apply`. Will not treat that as a blocker for the issue, as this needs to get in for data sanity during upgrade.

leelavg:
@amarts I'll take care of 'helm' when I'm back (mostly after this PR merge)

I guess we need rolling update for `server` pods as well, what do you feel?

If you don't mind can the PR be opened till Monday EOD?

doc/quick-start-yaml.adoc
```diff
@@ -5,6 +5,7 @@
 [source,console]
 ----
 $ kubectl apply -f https://github.com/kadalu/kadalu/releases/latest/download/kadalu-operator.yaml
+$ kubectl apply -f https://github.com/kadalu/kadalu/releases/latest/download/csi-nodeplugin.yaml
```
leelavg, line 8: I think, having a note here about why this is decoupled wrt upgrade and node drain would be good.

leelavg (commented):
LGTM

doc/quick-start-yaml.adoc
amarts: #608

PR #619, opened 2021-08-10
Fix external quota and expanding PVC

leelavg:

### DIFF
--- a/csi/controllerserver.py
+++ b/csi/controllerserver.py
@@ -36,6 +36,8 @@ def execute_gluster_quota_command(privkey, user, host, gvolname, path, size):
     """
     Function to execute the GlusterFS's quota command on external cluster
     """
+    # 'size' can always be parsed as integer with no errors
+    size = int(size) * 0.95
     quota_cmd = [
         "ssh",
         "-oStrictHostKeyChecking=no",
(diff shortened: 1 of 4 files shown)


leelavg:
@amarts, @aravindavk pls don't mind, I'm stressing it again.

- We should find a way including external gluster in our CI before we move further (or take up any enhancements)
- This is also a sneaking bug, wasn't able to find without dropping into `pdb` 😓 
- We aren't saving `pv_metadata` if using gluster-quota earlier, however expandvolume depends on that.

amarts (approved):

csi/controllerserver.py
```diff
@@ -49,7 +51,7 @@ def execute_gluster_quota_command(privkey, user, host, gvolname, path, size):
         "%s" % gvolname,
         "limit-usage",
         "/%s" % path,
-        "%s" % size * 0.05,
+        "%s" % str(size),
```
aravindavk: nit: `str` is not required here

leelavg: - ya, just wanted to be explicit (as it anyways will be turned to string).
- will remove it

aravindavk (approved):

PR #622, opened 2021-08-17
Added Prometheus metrics to expose the capacities of sub volumes

tili41:
Hi,
I couldn't get the gluster-metrics-exporter app to work in an internal glusterfs context, so I made a small improvement in csi/export.py to expose in Prometheus the disk capacities of sub volumes (free, used, total).

### DIFF
--- a/csi/exporter.py
+++ b/csi/exporter.py
@@ -7,7 +7,8 @@ from prometheus_client.core import GaugeMetricFamily, \
      CounterMetricFamily, REGISTRY
 from prometheus_client import start_http_server
 
-from volumeutils import (HOSTVOL_MOUNTDIR, PV_TYPE_SUBVOL)
+from volumeutils import (HOSTVOL_MOUNTDIR, PV_TYPE_SUBVOL,
+                         yield_pvc_from_mntdir)
 from kadalulib import logging_setup, logf
 
 
@@ -45,20 +46,20 @@ class CsiMetricsCollector(object):
             'Kadalu Storage Inodes free Count',
             labels=capacity_labels
         )
-        subvol_capacity_bytes = GaugeMetricFamily(
-            'kadalu_storage_subvol_capacity_bytes',
-            'Kadalu Storage Sub Volume Capacity',
-            labels=capacity_labels+["subvol"]
+        pv_capacity_bytes = GaugeMetricFamily(
+            'kadalu_storage_pv_capacity_bytes',
+            'Kadalu Storage PV Capacity',
+            labels=capacity_labels+["pv"]
         )
-        subvol_capacity_used_bytes = GaugeMetricFamily(
-            'kadalu_storage_subvol_capacity_used_bytes',
-            'Kadalu Storage Sub Volume Used Capacity',
-            labels=capacity_labels+["subvol"]
+        pv_capacity_used_bytes = GaugeMetricFamily(
+            'kadalu_storage_pv_capacity_used_bytes',
+            'Kadalu Storage PV Used Capacity',
+            labels=capacity_labels+["pv"]
         )
-        subvol_capacity_free_bytes = GaugeMetricFamily(
-            'kadalu_storage_subvol_capacity_free_bytes',
-            'Kadalu Storage Sub Volume Free Capacity',
-            labels=capacity_labels+["subvol"]
+        pv_capacity_free_bytes = GaugeMetricFamily(
+            'kadalu_storage_pv_capacity_free_bytes',
+            'Kadalu Storage PV Free Capacity',
+            labels=capacity_labels+["pv"]
         )
 
         for dirname in os.listdir(HOSTVOL_MOUNTDIR):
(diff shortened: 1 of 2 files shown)


csi/exporter.py
```diff
@@ -67,12 +83,33 @@ def collect(self):
...
+                        pvcname = os.path.basename(subvol)
+                        pvclabels = labels + [pvcname]
+                        try:
+                            # Get extended attributes for pvc
+                            size = os.getxattr(subvol, "trusted.gfs.squota.size")
+                            limit = os.getxattr(subvol, "trusted.gfs.squota.limit")
+                            subvol_capacity_bytes.add_metric(pvclabels, limit)
+                            subvol_capacity_used_bytes.add_metric(pvclabels, size)
+                            # Convert bytes type into int to compute the free capacity of the subvol
+                            free = int(limit.decode("utf-8")) - int(size.decode("utf-8"))
```
amarts: I feel there is a minor issue here. The `trusted.gfs.squota.size` is a value set in 'each' brick/storage pod. When more than one storage volume is given in the mount, this would be returning from one of the brick only. Hence, `os.statvfs()` is  better to call with pvc path to get both total size, and also proper usage.

amarts:
@tili41 Thanks for the PR, and welcome to contributors club of Kadalu :clap: :+1: Appreciate the effort spent on this.

csi/exporter.py
```diff
@@ -67,12 +83,33 @@ def collect(self):
                 inodes_free_count.add_metric(labels, free)
                 inodes_used_count.add_metric(labels, used)
 
+                # Gathers capacity metrics for each subvol
+                pvcpath = pathlib.Path(os.path.join(pth, PV_TYPE_SUBVOL)).glob("*/*/*")
```
leelavg: - If PVCs are small in number it's fine to use `glob`, however i feel it may not scale well.
- If possible, pls amend [this](https://github.com/kadalu/kadalu/blob/b8e9c5fc70a3c5e52b21d006505e483b23c9b281/csi/volumeutils.py#L1225) function, but if it's too much to take leave it.

csi/exporter.py
```diff
@@ -67,12 +83,33 @@ def collect(self):
...
+                            # Get extended attributes for pvc
+                            size = os.getxattr(subvol, "trusted.gfs.squota.size")
+                            limit = os.getxattr(subvol, "trusted.gfs.squota.limit")
+                            subvol_capacity_bytes.add_metric(pvclabels, limit)
+                            subvol_capacity_used_bytes.add_metric(pvclabels, size)
+                            # Convert bytes type into int to compute the free capacity of the subvol
+                            free = int(limit.decode("utf-8")) - int(size.decode("utf-8"))
+                            subvol_capacity_free_bytes.add_metric(pvclabels, free)
+                        except Exception:
+                            pass
```
leelavg: - Only `pass` an exception in the worst case
- Can you please log an error here?

leelavg:
@tili41 welcome and thanks for the PR.
- Can you pls sign the commit (`git commit -s`)?

csi/exporter.py
```diff
@@ -44,6 +45,21 @@ def collect(self):
             'Kadalu Storage Inodes free Count',
             labels=capacity_labels
         )
+        subvol_capacity_bytes = GaugeMetricFamily(
+            'kadalu_storage_subvol_capacity_bytes',
```
aravindavk: Instead of `_subvol_` change it to `_pv_` so that it will be familiar name. `subvol` is internal to Gluster/Kadalu

csi/exporter.py
aravindavk: +1 to `os.statvfs`

tili41: Yes indeed, this is a misunderstanding on my part. I didn't think that `os.statvfs` would get the glusterfs quotas at the pvc directory level.

csi/exporter.py
tili41: I used directly the `yield_pvc_from_mntdir` function since we are already in a loop, adding the `path_prefix` to the returned list.

csi/exporter.py
tili41: I deleted the code block that retrieved the xattr from the pvc, so it's not useful anymore.

tili41:
> @tili41 welcome and thanks for the PR.
> 
> * Can you pls sign the commit (`git commit -s`)?

csi/volumeutils.py
```diff
@@ -1219,7 +1219,7 @@ def yield_pvc_from_mntdir(mntdir):
                 data = json.loads(handle.read().strip())
             logging.debug(
                 logf("Found a PVC at", path=file_path, size=data.get("size")))
-            yield name[name.find("pvc"):name.find(".json")], data.get("size")
+            yield name[name.find("pvc"):name.find(".json")], data.get("size"), , data.get("path_prefix")
```
leelavg: - The extra comma (,) might be a syntax error, pls check
- You changed num of entries in return, earlier we are explicitly using 0 & 1 index, so this isn't a problem. Just mentioning it.

csi/exporter.py
leelavg: Yep, good call.

csi/volumeutils.py
tili41: Oops, sorry for the typo error

aravindavk (approved):

leelavg (approved):

PR #637, opened 2021-08-26
Use buildx while building local images

leelavg:
- Use buildx for building local images to simulate release scripts
- Break python instructions onto separate line from builder dockerfile
- Copy only requirements files while building local images

### DIFF
--- a/build.sh
+++ b/build.sh
@@ -2,14 +2,16 @@
 
 set -e -o pipefail
 
-cp requirements/* .
+cp -f requirements/*-requirements.txt .
 trap "rm -f *-requirements.txt" EXIT
 
 DOCKER_USER="${DOCKER_USER:-kadalu}"
 KADALU_VERSION="${KADALU_VERSION}"
 
 RUNTIME_CMD=${RUNTIME_CMD:-docker}
-build="build"
+# Use buildx for docker to simulate release script in github workflow
+# Requires Docker >=v19.03
+build="buildx build"
 if [[ "${RUNTIME_CMD}" == "buildah" ]]; then
         build="bud"
 fi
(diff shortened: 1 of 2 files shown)


leelavg:
- Not sure why we aren't using buildx from start, either it was experimental when project started or some other reason that I'm not aware of
- Please comment if there are reservations against using buildx by default
- We can go a step further and build multiarch (atleast arm64?) for only component images with buildx (<5min) and just run image with docker just to verify no dependency issues?

amarts:
* We didn't use buildx from start as in July 2019 it was not yet stable.
* Good to use it IMO
* I believe its not easy to run a different arch image ? How do you intend to test in CI/CD ?

amarts (approved):
LGTM

aravindavk (approved):

leelavg:
> I believe its not easy to run a different arch image ? How do you intend to test in CI/CD ?

well it's a long shot, here's what I'm thinking:
1. Use [setup-qemu-action](https://github.com/docker/setup-qemu-action) in every pr flow (<10s)
2. Edit `build.sh` to build (via buildx) arm64 variant of only component images (we'll pull/cache builder arm instance)
3. After amd64 tests complete, we'll run component images via docker (exact sha256 of an image can be used with docker to run arm images) just to see that the monitor process is starting in all images or not
4. When we are uncertain whether whole arm build is needed to verify a breaking change, we manipulate via commit message to build all containers

however this is not an easy task and as we are almost there unifying arm and amd images I intend to implement above in worst case.

Best way is, with this PR we are using buildx, before a release we'll use AWS machine to build containers and perform some sanity checks.

PR #639, opened 2021-08-30
Fix all possible docker issues wrt local and CI builds

leelavg:
Will try to explain in short and concise about the change:

Buildx:
1. Things are simpler if building on an amd64 platform for amd64 image
2. `--platform` and `--load` is must while building images on non amd64 host in `build.sh`
3. `--load` will not work if `--platform` has more than one entry
4. However, `--push` and `--load` work if `--platform` has only one entry and `--load` is used while building
dependent image in the next step to first pull from local docker
5. If `--platform` contains more than one entry, it's mandatory to use `--push`
6. Faulty release for 0.8.5 non amd64 is likely due to existing cache and `no-cache` is set for release CI
7. No way to verify point (6) unless we do an alpha with this PR

Fixes: #633

Signed-off-by: Leela Venkaiah G <leelavg@thoughtexpo.com>

PR #641, opened 2021-08-31
Use absolute pip path while building builder image

leelavg:
Ref: #633, #639

### DIFF
--- a/.github/workflows/on-pr-merge.yml
+++ b/.github/workflows/on-pr-merge.yml
@@ -74,7 +74,6 @@ jobs:
           platforms: linux/amd64
           target: prod
           push: true
-          load: true
           tags: |
             docker.io/kadalu/kadalu-csi:${{ github.sha }}
             docker.io/kadalu/kadalu-csi:devel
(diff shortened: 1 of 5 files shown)


leelavg:
- Release CI logs suggest only one thing
- By the time VIRTUAL_ENV is getting activated, `python3 -m pip` is installating packages
- And these pkgs are ending up in system python path
- Wait for a sec, log python, pip paths and use absolute pip path
- If virtualenv isn't activated absolute pip path will fail rather than
falling back to system's pip
- Still unsure why it's working on amd but not on arm

amarts (commented):
LGTM

The reason I introduced 'VIRTUAL_ENV'  was to make the process of copying packages from that of builder image easier. Considering we use python3-slim images now, I recommend removing VIRTUAL_ENV altogether, so it works seemless. See if removing VIRTUAL_ENV makes life any easier.

leelavg:
@amarts ack,

- The issue is too specific like, amd host + docker qemu + buildx with container driver + arm image
- Setting up this in local machine is non trivial so alpha releases are the only way
- Wrt venv, this is needed in builder image as we are installing a lot of python pkgs and they shouldn't be copied as it is to component images 
- To be frank, the current flow is very good and if this isn't going to work:
a. Will wait till I get a response on https://github.com/docker/build-push-action/issues/450 or
b. Will refactor builder image only for arm

amarts (approved):

PR #642, opened 2021-08-31
csi: Mounted Block Volume support 

aravindavk:
Create a Storage Class with `pv_type: Block` and use that
Storage Class to claim PVs.

Example Storage Class

```yaml
kind: StorageClass
apiVersion: storage.k8s.io/v1
metadata:
  name: kadalu.storage-pool-1-block
provisioner: kadalu
parameters:
  pv_type: Block
```

### DIFF
--- a/csi/controllerserver.py
+++ b/csi/controllerserver.py
@@ -13,7 +13,7 @@ import grpc
 from kadalulib import CommandException, logf, send_analytics_tracker, execute
 from volumeutils import (HOSTVOL_MOUNTDIR, PV_TYPE_SUBVOL, PV_TYPE_VIRTBLOCK,
                          check_external_volume, create_subdir_volume,
-                         create_virtblock_volume, delete_volume, expand_volume,
+                         create_block_volume, delete_volume, expand_volume,
                          get_pv_hosting_volumes, is_hosting_volume_free,
                          mount_and_select_hosting_volume, search_volume,
                          unmount_glusterfs, update_free_size,
@@ -70,6 +70,12 @@ def execute_gluster_quota_command(privkey, user, host, gvolname, path, size):
     return None
 
 
+def pvc_access_mode(request):
+    """Fetch Access modes from Volume capabilities"""
+    for vol_capability in request.volume_capabilities:
+        return vol_capability.access_mode.mode
+
+
 class ControllerServer(csi_pb2_grpc.ControllerServicer):
     """
     ControllerServer object is responsible for handling host
@@ -112,17 +118,24 @@ class ControllerServer(csi_pb2_grpc.ControllerServicer):
         pvsize = request.capacity_range.required_bytes
 
         pvtype = PV_TYPE_SUBVOL
-        # 'latest' finds a place here, because only till 0.5.0 version
-        # we had 'latest' as a separate version. After that, 'latest' is
-        # just a link to latest version.
-        if KADALU_VERSION in ["0.5.0", "0.4.0", "0.3.0"]:
-            for vol_capability in request.volume_capabilities:
-                # using getattr to avoid Pylint error
-                single_node_writer = getattr(csi_pb2.VolumeCapability.AccessMode,
-                                             "SINGLE_NODE_WRITER")
 
-                if vol_capability.access_mode.mode == single_node_writer:
-                    pvtype = PV_TYPE_VIRTBLOCK
+        # Mounted BlockVolume is requested via Storage Class.
+        # GlusterFS File Volume may not be useful for some workloads
+        # they can request for the Virtual Block formated and mounted
+        # as default MountVolume.
+        if request.parameters.get("pv_type", "").lower() == "block":
+            pvtype = PV_TYPE_VIRTBLOCK
+
+            single_node_writer = getattr(csi_pb2.VolumeCapability.AccessMode,
+                                         "SINGLE_NODE_WRITER")
+
+            # Multi node writer is not allowed for PV_TYPE_VIRTBLOCK
+            if pvc_access_mode(request) != single_node_writer:
+                errmsg = "Only SINGLE_NODE_WRITER is allowed for block Volume"
+                logging.error(errmsg)
+                context.set_details(errmsg)
+                context.set_code(grpc.StatusCode.INVALID_ARGUMENT)
+                return csi_pb2.CreateVolumeResponse()
 
         logging.debug(logf(
             "Found PV type",
@@ -235,8 +248,8 @@ class ControllerServer(csi_pb2_grpc.ControllerServicer):
                     return csi_pb2.CreateVolumeResponse()
 
                 if pvtype == PV_TYPE_VIRTBLOCK:
-                    vol = create_virtblock_volume(
-                        mntdir, request.name, pvsize)
+                    vol = create_block_volume(
+                        pvtype, mntdir, request.name, pvsize)
                 else:
                     use_gluster_quota = False
                     if (os.path.isfile("/etc/secret-volume/ssh-privatekey") \
@@ -335,8 +348,8 @@ class ControllerServer(csi_pb2_grpc.ControllerServicer):
 
         mntdir = os.path.join(HOSTVOL_MOUNTDIR, hostvol)
         if pvtype == PV_TYPE_VIRTBLOCK:
-            vol = create_virtblock_volume(
-                mntdir, request.name, pvsize)
+            vol = create_block_volume(
+                pvtype, mntdir, request.name, pvsize)
         else:
             use_gluster_quota = False
             vol = create_subdir_volume(
--- a/csi/volumeutils.py
+++ b/csi/volumeutils.py
@@ -20,6 +20,7 @@ GLUSTERFS_CMD = "/opt/sbin/glusterfs"
 MOUNT_CMD = "/bin/mount"
 UNMOUNT_CMD = "/bin/umount"
 MKFS_XFS_CMD = "/sbin/mkfs.xfs"
+XFS_GROWFS_CMD = "/sbin/xfs_growfs"
 RESERVED_SIZE_PERCENTAGE = 10
 HOSTVOL_MOUNTDIR = "/mnt"
 VOLFILES_DIR = "/kadalu/volfiles"
@@ -245,10 +246,10 @@ def mount_and_select_hosting_volume(pv_hosting_volumes, required_size):
     return None
 
 
-def create_virtblock_volume(hostvol_mnt, volname, size):
+def create_block_volume(pvtype, hostvol_mnt, volname, size):
     """Create virtual block volume"""
     volhash = get_volname_hash(volname)
-    volpath = get_volume_path(PV_TYPE_VIRTBLOCK, volhash, volname)
+    volpath = get_volume_path(pvtype, volhash, volname)
     volpath_full = os.path.join(hostvol_mnt, volpath)
     logging.debug(logf(
         "Volume hash",
@@ -261,7 +262,7 @@ def create_virtblock_volume(hostvol_mnt, volname, size):
     # Create a file with required size
     makedirs(os.path.dirname(volpath_full))
     logging.debug(logf(
-        "Created virtblock directory",
+        f"Created {pvtype} directory",
         path=os.path.dirname(volpath)
     ))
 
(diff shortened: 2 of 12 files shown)


aravindavk:
Still WIP. Remaining tasks are:

- Adding tests
- PV Expand fixes(Can be a future enhancement)
- Documentation changes

amarts (commented):
This looks pretty simple for the feature it provides :heart: 

Documentation would be key for this feature IMO. Great one to make many more applications use kadalu CSI now. After this gets merged, we can update the k8s documentation on csi drivers.

amarts:
```
E0831 14:49:28.575099       1 controller.go:966] error syncing claim "fb95480b-ac1f-49cc-886e-16de972ca683": failed to provision volume with StorageClass "kadalu.storage-pool-4": rpc error: code = InvalidArgument desc = Only SINGLE_NODE_WRITER is allowed for block Volume
```

This error is seen. Looks like RWO is default treated as block volume?

aravindavk:
> ```
> 
> E0831 14:49:28.575099       1 controller.go:966] error syncing claim "fb95480b-ac1f-49cc-886e-16de972ca683": failed to provision volume with StorageClass "kadalu.storage-pool-4": rpc error: code = InvalidArgument desc = Only SINGLE_NODE_WRITER is allowed for block Volume
> 
> ```
> 
> 
> 
> This error is seen. Looks like RWO is default treated as block volume?

I missed one extra check, I should compare rwo and virtblock.

leelavg:
I was awaiting @amarts to comment first 😅 

With first look, felt good wrt diff vs feature 😄, will review at the earliest.

Thanks.

aravindavk:
> I was awaiting @amarts to comment first 😅
> 
> With first look, felt good wrt diff vs feature 😄, will review at the earliest.
> 
> Thanks.

Thanks.

The diff is small because we had Virtblock feature earlier but disabled. Now that is reused with modifications required for raw Block Volume. :)

leelavg:
@aravindavk just had a look at test yaml files, we are no more using bare `provisioner: kadalu` [here](https://github.com/kadalu/kadalu/pull/642/files#diff-b6d8d3c489a8e870d58bb5324891bf73bb1d739f83a1cebe93627aae9a97eb6dR73) likewise in all sc's. Can you re-look at it once?

Edit: Incorrect.

amarts:
also, I believe for RawBlockStorage we need to publish the support? is there anything specific we need to do in capabilities?

aravindavk:
> @aravindavk just had a look at test yaml files, we are no more using bare `provisioner: kadalu` [here](https://github.com/kadalu/kadalu/pull/642/files#diff-b6d8d3c489a8e870d58bb5324891bf73bb1d739f83a1cebe93627aae9a97eb6dR73) likewise in all sc's. Can you re-look at it once?

Thanks. Copied from the doc page. I will fix that as well.

https://kadalu.io/docs/k8s-storage/devel/storage-classes/

aravindavk:
> also, I believe for RawBlockStorage we need to publish the support? is there anything specific we need to do in capabilities?

I think it is available. I see PVs are Bound. But Pods are failing.

leelavg:
> > also, I believe for RawBlockStorage we need to publish the support? is there anything specific we need to do in capabilities?
> 
> I think it is available. I see PVs are Bound. But Pods are failing.

I hope you are using `devicePath` (or something like that) in place of `mountPath`

aravindavk:
> > @aravindavk just had a look at test yaml files, we are no more using bare `provisioner: kadalu` [here](https://github.com/kadalu/kadalu/pull/642/files#diff-b6d8d3c489a8e870d58bb5324891bf73bb1d739f83a1cebe93627aae9a97eb6dR73) likewise in all sc's. Can you re-look at it once?
> 
> Thanks. Copied from the doc page. I will fix that as well.
> 
> https://kadalu.io/docs/k8s-storage/devel/storage-classes/

Just realized that, this is not the SC name. This is the CSI provisioner name that is registered, so looks correct.  https://github.com/kadalu/kadalu/blob/devel/templates/csi-driver-object.yaml.j2#L4

csi/controllerserver.py
```diff
@@ -70,6 +70,26 @@ def execute_gluster_quota_command(privkey, user, host, gvolname, path, size):
...
 
 
+def is_raw_block_pvc(request):
+    """If `volumeMode: Block` is specified in PVC"""
+    block = False
+    for vol_capability in request.volume_capabilities:
+        logging.info("HERE: ")
+        logging.info(vol_capability.access_type)
+        logging.info(csi_pb2.VolumeCapability.BlockVolume)
+        if vol_capability.access_type == csi_pb2.VolumeCapability.BlockVolume:
```
amarts:
```suggestion
        if vol_capability.get('access_type', '') == csi_pb2.VolumeCapability.BlockVolume:
```

would that resolve the problem? Looks like for block it is getting 'access_type as the option, but for everything else, its not having this parameter in argument, hence I suspect the failure is?

aravindavk: It is kind of different. "access_type" is `oneof` type in volume context. It can have either block or mount type.

```
// Specifies what API the volume will be accessed using. One of the
// following fields MUST be specified.
  oneof access_type {
    BlockVolume block = 1;
    MountVolume mount = 2;
  }
```

This doc suggests to use `WhichOneof` to find the name.

https://developers.google.com/protocol-buffers/docs/reference/python-generated#oneof

aravindavk:
One last issue remaining about mounting the raw block volume. Will check that tomorrow

```
 Type     Reason                  Age                  From                     Message
369
  ----     ------                  ----                 ----                     -------
370
  Warning  FailedScheduling        99s (x3 over 2m10s)  default-scheduler        0/1 nodes are available: 1 pod has unbound immediate PersistentVolumeClaims.
371
  Normal   Scheduled               94s                  default-scheduler        Successfully assigned default/pod3-4 to fv-az133-50
372
  Normal   SuccessfulAttachVolume  94s                  attachdetach-controller  AttachVolume.Attach succeeded for volume "pvc-928ae851-227c-48c4-ba85-8ac33c7b5e57"
373
  Warning  FailedMapVolume         28s (x8 over 93s)    kubelet                  MapVolume.EvalHostSymlinks failed for volume "pvc-928ae851-227c-48c4-ba85-8ac33c7b5e57" : lstat /var/lib/kubelet/plugins/kubernetes.io/csi/volumeDevices/publish/pvc-928ae851-227c-48c4-ba85-8ac33c7b5e57/dd9252bc-2b94-4d96-af12-f2be7a48398c: no such file or directory
```

@Madhu-1 any suggestions here? `target_path` is created while mounting. Not yet sure about the error.

leelavg:
@aravindavk and @amarts  just connecting the dots and below are my observations, still didn't review the code 😓, anyways:

-  `pv_type: Block` (aka virtblock) is working fine without issues
-  `volumeMode: Block` (aka RawBlock) is almost working and failing when k8s tries to stat on a file which it assumes to exist in that specific path

Observations:
- Mostly the bug is at creating only mountpoint/parent-dir but not the file [here](https://github.com/kadalu/kadalu/pull/642/files#diff-850490ad0b2666d1ea49b540576d0e6bb6525baf9e7b87d84fd841f80758ec4eR824-R825)
- From log we say we created this file, however we are creating only the parent directory but CO assumes a device file there
```
-> What we say we did
2021-09-01T16:00:31.4250656Z [2021-09-01 15:57:47,577] INFO [nodeserver - 113:NodePublishVolume] - Mounted PV	 volume=pvc-b83dc0bb-4484-4745-84b0-4bba45d79b8c pvpath=rawblock/c3/db/pvc-b83dc0bb-4484-4745-84b0-4bba45d79b8c pvtype=rawblock hostvol=storage-pool-4 target_path=/var/lib/kubelet/plugins/kubernetes.io/csi/volumeDevices/publish/pvc-b83dc0bb-4484-4745-84b0-4bba45d79b8c/4b149164-0b9c-4207-8238-05f1313f41b2 duration_seconds=0.013628005981445312
-> What CO is assuming
2021-09-01T15:59:38.7255448Z   Warning  FailedMapVolume         47s (x8 over 112s)     kubelet                  MapVolume.EvalHostSymlinks failed for volume "pvc-b83dc0bb-4484-4745-84b0-4bba45d79b8c" : lstat /var/lib/kubelet/plugins/kubernetes.io/csi/volumeDevices/publish/pvc-b83dc0bb-4484-4745-84b0-4bba45d79b8c/4b149164-0b9c-4207-8238-05f1313f41b2: no such file or directory
```
- It's failing when performing `lstat` on the file. Quoting directly from [csi-doc](https://kubernetes-csi.github.io/docs/raw-block.html)
> The main difference between block volumes and mount volumes is the expected result of the NodePublish(). For mount volumes, the CO expects the result to be a mounted directory, at TargetPath. For block volumes, the CO expects there to be a device file at TargetPath. The device file can by a bind-mounted device from the hosts /dev file system, or it can be a device node created at that location using mknod().

- So the solution might be, use `mknod()` to create the device file or something of that sort 😄

csi/volumeutils.py
```diff
@@ -804,22 +809,32 @@ def volume_list(voltype=None):
...
 
 def mount_volume(pvpath, mountpoint, pvtype, fstype=None):
     """Mount a Volume"""
     # Need this after kube 1.20.0
-    makedirs(mountpoint)
+    if pvtype == PV_TYPE_RAWBLOCK:
+        # Only create the parent directory since the Target
+        # is raw device file.
+        makedirs(os.path.dirname(mountpoint))
+        Path(mountpoint).touch(mode=0o660)
```
Madhu-1: we need to create a file with provided targetDirectory. if this is doing that then we need to bind mount the stagingPath to the targetPath.

aravindavk: `NodeStageVolume` is not implemented in Kadalu CSI, File from Gluster Vol is directly bind mounted as `target_path` as file. Let me explore about `stagingPath`

aravindavk:
Now trying to do `losetup` to the staging path and then bind mount to target Path.

examples/sample-test-app1.yaml
```diff
@@ -65,3 +65,76 @@ spec:
...
+      claimName: pv1-3
+  restartPolicy: OnFailure
+---
+apiVersion: v1
+kind: PersistentVolumeClaim
+metadata:
+  name: pv1-4
+spec:
+  accessModes:
+    - ReadWriteMany
```
amarts:
```suggestion
    - ReadWriteOnce
```

examples/sample-test-app2.yaml
```diff
@@ -66,3 +66,76 @@ spec:
...
+      claimName: pv2-3
+  restartPolicy: OnFailure
+---
+apiVersion: v1
+kind: PersistentVolumeClaim
+metadata:
+  name: pv2-4
+spec:
+  accessModes:
+    - ReadWriteMany
```
amarts:
```suggestion
    - ReadWriteOnce
```

examples/sample-test-app3.yaml
```diff
@@ -66,3 +66,76 @@ spec:
...
+      claimName: pv3-3
+  restartPolicy: OnFailure
+---
+apiVersion: v1
+kind: PersistentVolumeClaim
+metadata:
+  name: pv3-4
+spec:
+  accessModes:
+    - ReadWriteMany
```
amarts:
```suggestion
    - ReadWriteOnce
```

amarts (commented):
Isn't all block storage have RWO mode of operation?

amarts:
aravinda, it looks like one part of the block storage (rawblock) is working fine now. Should we just promote that part and take the PR in? That way, we can allow some more testing with 0.8.6 (as we need to quickly make a release for ARM64 fixes). And fix the virtblock storage thing in eventual releases?

aravindavk:
> aravinda, it looks like one part of the block storage (rawblock) is working fine now. Should we just promote that part and take the PR in? That way, we can allow some more testing with 0.8.6 (as we need to quickly make a release for ARM64 fixes). And fix the virtblock storage thing in eventual releases?

Makes sense. I will split the PR into two, one to enable Mounted Block Volume support(Working now) and other one as raw Block volume support.

aravindavk:
@amarts Updated the PR with only Mounted Block Volume feature. I will send new PR for Raw Block Volume.

amarts (approved):

amarts:
Looks like we need to increase the timeout by another 30 seconds as one more test is added.

PR #645, opened 2021-09-04
[Obsolete] Remount PVC on nodeplugin reboot

leelavg:
Approach:
1. Store minimal data on storage pool to be able to reconstruct target_path on correct node upon nodeplugin reboot
2. Once the nodeplugin reboots, mount all storage pool and search for all PVCs which has target_path
3. Cleanup/unmount damaged mount point and (bind) mount the PVC

Pros:
1. Automatically remount volumes of type 'subdir/virtblock && native' on nodeplugin reboot
2. Separate yaml for operator and nodeplugin isn't needed (it's unaviodable even with separate yaml to not have downtime)

Cons:
1. On nodeplugin reboot, startup time may be noticeable (~2min) before serving requests
2. Extra storage space used on storage pool (increases linearly wrt num of target_paths)
3. Couldn't remount non_native volumes as of now, maybe doable with other approaches

Unknowns:
1. Metadata heal due to usage of `truncate` while updating metadata file?
2. Need to add logic when rawBlock is supported

Alternates (one of):
1. Use kubectl in csi pods as well and query for PVC metadata directly rather than using storage pool
2. Increase the use of k8s python api (it's very less as of now) to get any resource/kind data programatically
3. Create endpoints in operator and serve required data for all other components

### DIFF
--- a/csi/volumeutils.py
+++ b/csi/volumeutils.py
@@ -865,9 +865,9 @@ def unmount_glusterfs(mountpoint):
         execute(UNMOUNT_CMD, "-l", mountpoint)
 
 
-def unmount_volume(mountpoint):
+def unmount_volume(mountpoint, force=False):
     """Unmount a Volume"""
-    if os.path.ismount(mountpoint):
+    if os.path.ismount(mountpoint) or force:
         execute(UNMOUNT_CMD, "-l", mountpoint)
 
 
@@ -1235,10 +1235,6 @@ def remount_storage():
     Mount storage if any volumes exist after a pod reboot
     """
     host_volumes = get_pv_hosting_volumes({})
-    all_pools = {
-        os.path.join(HOSTVOL_MOUNTDIR, volume["name"])
-        for volume in host_volumes
-    }
     if os.environ.get("CSI_ROLE", "-") == "provisioner":
         for volume in host_volumes:
             if volume["kformat"] == "non-native":
@@ -1254,7 +1250,6 @@ def remount_storage():
                 logging.error(
                     logf("Unable to mount volume", hvol=hvol, err=err))
     elif os.environ.get("CSI_ROLE", "-") == "nodeplugin":
-        mounted_pools = set()
         for pvc in yield_pvc_from_hostvol():
             if not (pvc.get("mounts") and pvc.get("mounts").get(NODE_ID)):
                 continue
@@ -1268,9 +1263,14 @@ def remount_storage():
             for target in pvc.get("mounts").get(NODE_ID):
                 try:
                     # TODO: Remount non-native PVCs
+                    # Upon nodeplugin reboot PVC mounts go stale and first need
+                    # to be fixed/unmounted
+                    if not os.path.ismount(target):
+                        unmount_volume(target, force=True)
                     mount_volume(pvpath_full, target, pvtype)
-                    mounted_pools.add(pvc["mntdir"])
-                    # Fill the cache, not to faill if CO calls NodeUnpublishVolume
+
+                    # Fill the cache, not to faill if CO calls
+                    # NodeUnpublishVolume
                     PVC_POOL[target] = (pvc["mntdir"],
                                         os.path.join(pvc["path_prefix"],
                                                      pvc["name"]))
@@ -1284,10 +1284,6 @@ def remount_storage():
                              path=pvpath_full,
                              target=target,
                              error=err))
-        # Unmount pools which doesn't have a PVC mounted on this node
-        for pool in all_pools - mounted_pools:
-            unmount_glusterfs(pool)
-
 
 # Methods starting with 'yield_*' upon known exceptions/empty returns are
 # designed to yield None. Caller should handle None gracefully based on the
@@ -1355,7 +1351,7 @@ def yield_pvc_from_hostvol():
             for data in pvc:
                 if data is not None:
                     pvc_exist = True
-                    data["mntdir"] = mntdir
+                    data["mntdir"] = os.path.dirname(mntdir.strip("/"))
                     yield data
     if not pvc_exist:
         # Empty yield if no PVC exist in any pool
--- a/csi/volumeutils.py
+++ b/csi/volumeutils.py
@@ -31,6 +31,10 @@ VOLINFO_DIR = "/var/lib/gluster"
 # Used while sending SIGHUP during any modifcation to storage config
 VOL_DATA = {}
 
+# Cpython dict is threadsafe and can be re-used across grpc calls
+# Used to store, target_path to tuple of mntdir, pvpath
+PVC_POOL = {}
+
 statfile_lock = threading.Lock()    # noqa # pylint: disable=invalid-name
 mount_lock = threading.Lock()    # noqa # pylint: disable=invalid-name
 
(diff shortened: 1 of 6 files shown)


leelavg:
- Didn't implement fully yet, will revisit and complete at the earliest 😄

csi/volumeutils.py
```diff
@@ -1196,6 +1200,31 @@ def check_external_volume(pv_request, host_volumes):
...
+            try:
+                mount_glusterfs(volume, mntdir)
+                logging.info(logf("Volume is mounted successfully", hvol=hvol))
+            except CommandException:
+                logging.error(logf("Unable to mount volume", hvol=hvol))
+    elif os.environ.get("CSI_ROLE", "-") == "nodeplugin":
+        # TODO:
+        # 1. Mount PVCs by utilizing target_path in pv metadata and
+        # comparing with existing path for Pod mount space
+        # 2. Refactor yield methods for correct usage
```
leelavg: Just a reminder to me.

3. Fill up PVC_POOL dict
4. Implement minimal health check that CSI is ready 
5. Name will not always start with https://github.com/kadalu/kadalu/blob/382c527ea4e836d3e088ba17da79a94a65a24fea/csi/volumeutils.py#L1233
6. Mounting a RWX pvc?

leelavg:
@aravindavk , @amarts before making any changes or fixing failures in this PR, just wanted to check whether you approve the approach or not:

In short:
1. Store target_path and other minimal info directly in \<pvc-volume\>.json file and in a dict cache during NodePublish
2. During NodeUnpublish construct pool data from dict cache and remove target_path entry in \<pvc-volume\>.json
3. If nodeplugin is rebooted, we can construct target_path to mount PVC and warm up the dict cache to be used in NodeUnpublish

Will update the commit message with all the pros, cons, alternative, minor fixes etc, above is the basic idea.

leelavg:
- Working but I felt this is hacky with unwanted side effects
- Just had a re-look at proto def, all we need is to implement `NodeStage` and `NodeUnstage` volume
- Rationale is when NodePlugin is rebooted, CO will re-initiate* NodePublish where we'll get only target_path, but `NodeStage` gets whole publish context from `CreateVolume` and that's what this PR is trying to persist on gluster pool

\* only an observation, need to spend time to confirm this

aravindavk:
> Rationale is when NodePlugin is rebooted, CO will re-initiate* NodePublish where we'll get only target_path, but NodeStage gets whole publish context from CreateVolume and that's what this PR is trying to persist on gluster pool

I think we need to implement NodeStage and NodeUnstage. It is also required for raw block support.

leelavg:
- Not sure what I'm missing, may need more eyes
```
Before:
/mnt/sample-pvc # ls | wc -l
400

/mnt/sample-pvc # find . -type f -exec md5sum {} + | awk '{print $1}' | sort | md5sum
bfe25c32fd222f4ad3ea3d4719def85b  -

During nodeplugin reboot:
# k delete po kadalu-csi-nodeplugin-96rfd --force
warning: Immediate deletion does not wait for confirmation that the running resource has been terminated. The resource may continue to run on the cluster indefinitely.
pod "kadalu-csi-nodeplugin-96rfd" force deleted

/mnt/sample-pvc # ls
ls: .: Transport endpoint is not connected

# k logs kadalu-csi-nodeplugin-nkwcr -c kadalu-nodeplugin -f
[2021-09-07 12:43:26,778] INFO [kadalulib - 368:start_all] - Started Process     name=csi
[2021-09-07 12:43:26,782] INFO [kadalulib - 368:start_all] - Started Process     name=metrics
[2021-09-07 12:43:26,786] INFO [kadalulib - 368:start_all] - Started Process     name=volumewatch
Starting watch on configmap
Setting up watches.
Watches established.
[2021-09-07 12:43:27,510] INFO [exporter - 121:<module>] - Started Kadalu Storage CSI Metrics exporter.  port=8000
[2021-09-07 12:43:27,656] INFO [volumeutils - 1307:yield_hostvol_mount] - Volume is mounted successfully         hvol=sample-pool
[2021-09-07 12:43:27,698] DEBUG [volumeutils - 1335:yield_pvc_from_mntdir] - Found a PVC at      path=/mnt/sample-pool/info/subvol/9b/22/pvc-15af3a5e-2739-4164-b0a1-702bec19425d.json size=1073741824
[2021-09-07 12:43:27,727] INFO [volumeutils - 1277:remount_storage] - PVC is re-mounted successfully     pvc=pvc-15af3a5e-2739-4164-b0a1-702bec19425d target_path=/var/lib/kubelet/pods/2cf7b8cb-cea3-430e-bcf9-d7720fcd9794/volumes/kubernetes.io~csi/pvc-15af3a5e-2739-4164-b0a1-702bec19425d/mount
[2021-09-07 12:43:27,748] INFO [main - 54:main] - Server started

After reboot:
/mnt/sample-pvc # ls
ls: .: Transport endpoint is not connected
```
- Maybe I need to re-mount `/var/lib/kubelet/plugins/kubernetes.io/csi/pv/<pvc>/globalmount` as well 🤔

leelavg:
@aravindavk 
> I think we need to implement NodeStage and NodeUnstage. It is also required for raw block support.

Ack, IMHO in this case it's better to have either virtblock or rawBlock (preferred), atleast I'm a bit confused what one provides than the other. 

I'll break this PR into multiple PRs.

aravindavk:
> @aravindavk
> 
> > I think we need to implement NodeStage and NodeUnstage. It is also required for raw block support.
> 
> Ack, IMHO in this case it's better to have either virtblock or rawBlock (preferred), atleast I'm a bit confused what one provides than the other.

**Mounted Block Volume**: Filesystem created by the Provisioner itself. Applications gets the mounted volume. Good for some workloads since Gluster handles the I/O as single file Operation.
**Raw Block Volume**: Raw device available for the application. Application can choose the filesystem as required and mount it. 

Functionality wise both are similar, but Raw block volume gets more flexibility.

> 
> I'll break this PR into multiple PRs.

leelavg:
1. Part1 is separated into #647 
2. Part2 is separated into #658

leelavg:
- No alternatives found

sky-philipalmeida:
You can do a **PostStart** hook on **nodeplugin**(tested on version 1.1.0).
Everytime a nodeplugin restarts the mount points will be recreated, for that we use the following instruction:

`
mount | grep glusterfs | awk '{print $1 " " $3}' | sed 's/://' | awk '{print "umount " $3 ";glusterfs --process-name fuse --volfile-id " $2 " --volfile-server " $1 "--log-level=DEBUG " $3}' > /tmp/remount.sh ; chmod +x /tmp/remount.sh ; /tmp/remount.sh
`

On the **nodeplugin** deployment:

> 
          lifecycle:
            postStart:
              exec:
                command:
                  - /bin/sh
                  - '-c'
                  - >
                    mount | grep glusterfs | awk '{print $1 " " $3}' | sed 's/:/
                    /' | awk '{print "umount " $3 ";glusterfs --process-name
                    fuse --volfile-id " $2 " --volfile-server " $1 "
                    --log-level=DEBUG " $3}' > /tmp/remount.sh ; chmod +x
                    /tmp/remount.sh ; /tmp/remount.sh

**Important:** The workloads that use Kadalu volumes should also have a **liveness** check for the mounts (a simple `ls volume` is ok).
They will be forced to restart once the volume becomes unavailable but will be able to do it gracefully bypassing errors similar to:

`Error: failed to start container "imp-sentry-clickhouse": Error response from daemon: error while creating mount source path '/var/lib/kubelet/pods/654d3248-a244-4cc5-a7cb-9b50e4da2773/volumes/kubernetes.io~csi/pvc-44f6b9bd-6596-4d87-9be4-155e13a0f878/mount': mkdir /var/lib/kubelet/pods/654d3248-a244-4cc5-a7cb-9b50e4da2773/volumes/kubernetes.io~csi/pvc-44f6b9bd-6596-4d87-9be4-155e13a0f878/mount: file exists `

Issue #660, opened 2021-09-22
kadalu consume a lot of CPU


vincentchong1:
Will like to find out what is the reason kadalu is consuming a lot of CPU on the k8s worker node.

kadalu 0.8.4
Kubernetes v1.20.6
centos OS 3.10.0-1160.el7.x86_64

kubectl get pod -n kadalu -o wide
NAME                             READY   STATUS    RESTARTS   AGE   IP              NODE          NOMINATED NODE   READINESS GATES
kadalu-csi-nodeplugin-5wsmk      3/3     Running   0          39h   172.16.98.132   c432-kwo-02   <none>           <none>
kadalu-csi-nodeplugin-tnwld      3/3     Running   0          39h   172.16.162.69   c432-kwo-01   <none>           <none>
kadalu-csi-provisioner-0         5/5     Running   0          39h   172.16.98.131   c432-kwo-02   <none>           <none>
operator-56898988c5-tlcq6        1/1     Running   0          39h   172.16.162.65   c432-kwo-01   <none>           <none>
server-kadalu-storage-pool-0-0   1/1     Running   0          39h   172.16.162.68   c432-kwo-01   <none>           <none>
server-kadalu-storage-pool-1-0   1/1     Running   0          39h   172.16.98.130   c432-kwo-02   <none>           <none>

very high usage on this worker, kadalu/main.py is hover around 50%
[centos@C432-KWO-01 ~]$ ps -eo pid,ppid,cmd,%mem,%cpu --sort=-%cpu | head
  PID  PPID CMD                         %MEM %CPU
28618  6434 python3 /kadalu/main.py      0.1 49.7
 9315  8624 /opt/sbin/glusterfsd -N --v  0.0 19.9
13379     1 /usr/bin/kubelet --bootstra  0.6 14.8
13502  9290 /opt/sbin/glusterfs --proce  0.0 12.1
12894     1 /usr/bin/dockerd -H fd:// -  0.5  8.9
    1     0 /usr/lib/systemd/systemd --  0.0  7.0
20537     1 /usr/bin/perl -w /usr/bin/s  0.0  6.6
 4647  3076 /usr/local/bin/node /opt/tc  0.2  5.4
18358 18341 redis-server *:6379          0.0  4.9

but it is reasonable on this worker
[centos@C432-KWO-02 ~]$ ps -eo pid,ppid,cmd,%mem,%cpu --sort=-%cpu | head
  PID  PPID CMD                         %MEM %CPU
[...]

PR #661, opened 2021-09-24
Access k8s APIServer behind a proxy

leelavg:
Ref: #660


leelavg:
- Using default as `HTTP_PROXY` string isn't looking good, it is what it is in current setting
- Need to wait for user confirmation whether there's a proxy or not
- From logs it seems so but ya, confirmation doesn't hurt

Fix is from: https://github.com/kubernetes-client/python/issues/333
Similar Issue: https://github.com/kubernetes-client/python/issues/549

PR #670, opened 2021-10-16
handle the setfattr failure due to EPERM

amarts:
The failure to set the quota usage happens when its done on
non-namespace directory. Considering this is happening only on
PVC dirs, this fix of trying to set the 'namespace' xattr again
seems to be a best solution right now.

### DIFF
--- a/csi/quota-crawler.sh
+++ b/csi/quota-crawler.sh
@@ -21,7 +21,13 @@ while true; do
     used_size=$(df -B1 ${dir}  | tail -n1 | awk '{print $3}');
     out=$(setfattr -n glusterfs.quota.total-usage -v ${used_size} $dir 2>&1);
     if ! [[ "$out" =~ "Operation not supported" ]]; then
-      echo "$out"
+	echo "$out"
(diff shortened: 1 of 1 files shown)


aravindavk (approved):

csi/quota-crawler.sh
```diff
@@ -21,7 +21,13 @@ while true; do
...
     out=$(setfattr -n glusterfs.quota.total-usage -v ${used_size} $dir 2>&1);
     if ! [[ "$out" =~ "Operation not supported" ]]; then
-      echo "$out"
+	echo "$out"
+	# This error comes most probably after a server pod
+	# restart or after a self-heal. This is a quick fix.
+	# Ideal fix is handling it in glusterfs code, but
+	# that would take time. In worst case, it would be
+	# a no-op as namespace would be already set.
+	out1=$(setfattr -n trusted.glusterfs.namespace -v "true" $dir 2>&1);
```
leelavg, line 30: Since we aren't using `out1` better not to store anything?

PR #671, opened 2021-10-16
Document xfsprogs version requirements

gnossen:
While testing out the new VirtBlock feature, I initially saw failed mount operations. The Kubelet log contained:

```
mount: /var/lib/kubelet/pods/bfac4377-53ed-4f56-a09b-43bb35ae3437/volumes/kubernetes.io~csi/pvc-8f01e993-0bb0-4960-b564-9fbeb366748f/mount: wrong fs type, bad option, bad superblock on /dev/loop0, missing codepage or helper program, or other error."
```

Digging into things further, I found that the `xfsprogs` package installed on my host machines did not support the superblock v5 format, [as documented here](https://docs.oracle.com/en/operating-systems/oracle-linux/6/relnotes6.9/ol6-rn-issues-xfs-superblock-compat.html). By upgrading the distro on my host machines (and thereby the `xfsprogs` package), I was able to successfully mount VirtBlock PVs.

I wanted to supply stricter version guidance here, but it seems that Redhat-derived distros and Debian-derived distros number their `xfsprogs` packages differently and there doesn't seem to be a way to get consistent version information out of the various CLI utilities without creating an FS and noting what superblock version it used.

### DIFF
--- a/doc/setup.adoc
+++ b/doc/setup.adoc
@@ -3,7 +3,7 @@
 == Pre-Requisites
 
 - Kubernetes 1.13.0 + version
-- The host should support xfs (`mkfs.xfs`)
+- The host should support xfs (`mkfs.xfs`) with the superblock v5 format
   - On some systems this might require installation of xfsprogs package
 
 == Setup


aravindavk (commented):
Looks good. Any flags available with mkfs.xfs to not use this feature? We can use the same in code.

doc/setup.adoc
```diff
@@ -3,7 +3,7 @@
 == Pre-Requisites
 
 - Kubernetes 1.13.0 + version
-- The host should support xfs (`mkfs.xfs`)
+- The host should support xfs (`mkfs.xfs`) with the superblock v5 format
```
leelavg, line 6: Sry, piggybacking on this PR, can you also add, node should've loop device if virtblock pv is required 😉?

leelavg:
> Digging into things further, I found that the xfsprogs package installed on my host machines did not support the superblock v5 format

It is highly recommended to create xfs from host, in that scenario afaik this behaviour will not be seen.

Even after creating fs on disk, device option in manifest would still work, since we try to mount on empty disk first and fall back to creating xfs 😄

amarts:
>  we try to mount on empty disk first and fall back to creating xfs smile

@leelavg This case in 'VirtBlock' is different. It is a file inside glusterfs which is already created and started. So, the xfsprog is called from inside CSI itself (not operator)..

doc/setup.adoc
amarts: @leelavg last I saw, most of the Distros had support for this. Yes I agree that with RHEL 7.x the number of loop devices were less, and it needed manual creates.

leelavg:
@amarts yep, missed that detail. Thanks.

amarts:
By itself this PR is better than earlier. Even though there are comments, having this merged before 0.8.7 is good IMO. Hence merging.

PR #677, opened 2021-10-28
Use gluster-opts while mounting external vol

leelavg:
Ref: #562


amarts (approved):
LGTM. Better to have it and then fix any other issues which arises by using it.

Issue #678, opened 2021-10-28
kubectl kadalu healinfo fails with error code 234


RomanKisilenko:
I have kadalu 0.8.6 installed on a kubernetes 1.18.11 cluster where I configured a replica 1 volume. When I try to execute healinfo command, I'm getting:
```
$ kubectl kadalu healinfo
Error while running the following command
$ kubectl exec -nkadalu kadalu-csi-provisioner-0 -c kadalu-provisioner -- /kadalu/heal-info.sh

command terminated with exit code 234
```

It seems like there are no clues in kadalu logs.

Curious how this can be resolved.

hunter86bg:
Can you provide the output of the script:
```
kubectl exec -n kadalu -it kadalu-csi-provisioner-0 -c kadalu-provisioner -- bash
```
inside it 
```
bash -x /kadalu/heal-info.sh
```

hunter86bg:
Usually replica1 volumes do not heal to anywhere, so the output should be 'heal is not supported'

RomanKisilenko:
```
root@kadalu-csi-provisioner-0:/# bash -x /kadalu/heal-info.sh
+ MOUNT_DIR=/mnt
+ '[' 0 -eq 1 ']'
+ '[' 0 -gt 1 ']'
++ /bin/ls -d /mnt/replica1
++ wc -l
+ dirs=1
+ '[' 1 -lt 1 ']'
++ cd /kadalu/volfiles
++ ls replica1.client.vol
+ for volfile in $(cd /kadalu/volfiles; ls *)
+ vol=replica1
+ echo 'Giving heal information of volume replica1'
Giving heal information of volume replica1
+ /opt/libexec/glusterfs/glfsheal replica1 info-summary volfile-path /kadalu/volfiles/replica1.client.vol
This command is supported for only volumes of replicate/disperse type. Volume replica1 is not of type replicate/disperse
```

PR #683, opened 2021-11-01
Add spacing between volumes' output, fixes kadalu/kadalu#678

hunter86bg:


hunter86bg:
@amarts , @leelavg ,
can you take a look in this "change" .
It seems that when the 'kubectl-kadalu healinfo' is triggered and there is only a replica1 volume - the output from glfsheal is truncated by k8s due to the exit code. The echo will both space the output between 2 volumes (tested with replica1 + replica2) and also will return a '0' so the glfsheal output will be shown (that replica1 volumes cannot be healed).

amarts (approved):

leelavg:
- Thanks, this explanation makes sense. 
- LGTM

leelavg (approved):

aravindavk (approved):

Issue #684, opened 2021-11-02
self-heal Improvement - Provide list of all files which needs healing


amarts:
Currently, the CLI provided for self-heal listing only provides number of files which need self-heal on a volume.

We need to expand it to support all the commands available with gluster CLI, ie, to list individual files (or gfid) which are pending heal to help users fast track their healing.

Also there should be a command to force healing of a file (again something which is already present in gluster CLI)

amarts:
An example of triggering  a self-heal on complete volume : https://github.com/kadalu/kadalu/issues/676#issuecomment-957127152

PR #689, opened 2021-11-05
Add extendend heal info , fixes kadalu/kadalu#684

hunter86bg:
It seems that we need extended info only when total entries from info-summary is greater than 0 and is not "-" (replica 2).


hunter86bg:
It's strange that the indentation is changing (maybe I used 'tab's)

PR #693, opened 2021-11-08
Add support for `volumeMode: Block`

leelavg:

### DIFF
--- a/examples/sample-test-app1.yaml
+++ b/examples/sample-test-app1.yaml
@@ -69,7 +69,7 @@ spec:
 kind: StorageClass
 apiVersion: storage.k8s.io/v1
 metadata:
-  name: kadalu.storage-pool-1-block
+  name: kadalu.storage-pool-1-virtblock
 provisioner: kadalu
 parameters:
   storage_name: "storage-pool-1"
@@ -80,7 +80,7 @@ apiVersion: v1
 metadata:
   name: pv1-3
 spec:
-  storageClassName: kadalu.storage-pool-1-block
+  storageClassName: kadalu.storage-pool-1-virtblock
   accessModes:
     - ReadWriteOnce
   resources:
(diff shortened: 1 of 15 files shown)


leelavg:
~- Pls excuse the symlink for `kadalulib.py`, the side-effect is only re-order of import statements~
~- This suppresses the wrong diagnostics provided by LSP~
- I'll add docs after getting a pass run from CI

amarts:
can run a rebase now!

examples/sample-test-app1.yaml
```diff
@@ -106,3 +106,36 @@ spec:
...
+apiVersion: v1
+kind: Pod
+metadata:
+  name: pod1-4
+spec:
+  containers:
+    - name: bash-block-vol
+      image: bash
+      command: ["/usr/local/bin/bash","-c"]
+      args: ["echo sample text > /tmp/original && dd if=/tmp/original of=/dev/xvda && dd if=/dev/xvda of=/tmp/copy bs=1 count=12 && cat /tmp/{original,copy} && diff <(cat /tmp/original) <(cat /tmp/copy) && echo all good"]
```
aravindavk, line 132: @amarts Is this test sufficient to say it is actually a device file or a test required to create a fs and mount?

leelavg: - I'm facing issues in my local setup and on CI as well for creating and mounting a file system on this PV disk
- Docker in docker or similar setups seems not to be working quite well with mount namespace isolation/permissions
- My thought process for docs is
1. Use `pv_type: Block` if user wants to have a fs mounted on block file/device
2. Use `volume_mode: Block` if user want to have access to a block file/device

amarts: > Use volume_mode: Block if user want to have access to a block file/device

What are the usecases for this? Ie, a block device without mount.

leelavg: - I'm thinking of databases which can use raw disk but not quite sure though
- Nevertheless, if we want to be sure, we need to deploy containers created from this PR on actual node

amarts:
I am OK with having this piece merged as a Beta feature, and let users decide where it is better used. We can improve further later based on feedback.

amarts (approved):
@aravindavk when you feel like, merge it. (pre 0.8.8 or post)

aravindavk (approved):

PR #695, opened 2021-11-09
Symlink files for LSP diagnostics to work correctly

leelavg:
- Two symlinks at root of repo to make lsp work
- Re-order of imports is the side effect


leelavg:
- Will send second part of #693 after this PR is merged

amarts (approved):

PR #713, opened 2021-11-25
csi: Add support for storage-options through SC

### LINKED ISSUE #680: [RFE] Storage Pool Options(Volume Options) support
Storage class is the best way to adjust the behaviour of the PVs. This proposal talks about adding support for Storage pool options(Volume Options) to the Storage Class.

```
kind: StorageClass
apiVersion: storage.k8s.io/v1
metadata:
  name: kadalu.storage-pool-1
provisioner: kadalu
parameters:
  storage_options:
    performance.open-behind: on
    performance.write-behind: off
    performance.stat-prefetch: off
    performance.quick-read: off
    performance.strict-o-direct: on
    performance.read-ahead: off
    performance.io-cache: off
    performance.readdir-ahead: off
    performance.client-io-threads: on
    server.event-threads: 4
    client.event-threads: 4
    performance.read-after-open: yes
```

**Note**: The Option names added here is only for illustration purpose. Option names may change based on supported option names in Kadalu.

Changes:

- If Storage Class options not available in Node plugin(?), then we may have to add these options in Volume Context.(TBD)
- No change to Storage pool mount in provisioner
- Volgen function should accept the list of Options and uses them while generating Client Volfiles
- Node plugin mounts: Add hash to mount path and volfile path while mounting the Kadalu Storage(Host Volume)

```
hash = xxhash(storage_options_sorted.stringify)
mount_path = "/mnt/<volname>_<hash>
volfile_path = "<volfiles_dir>/<volname>.client_<hash>.vol"
```

I will add more details in the comments about using these Volume options to generate custom volfiles. Rest of the changes can be implemented independently assuming the volgen function takes care of using them.

Alternatively, Volume options can be configured while creating the Storage pool itself. Addition to the above changes, we also need changes to CRD, CLI and Operator.

vatsa287:
Fixes: #680

### DIFF
--- a/csi/controllerserver.py
+++ b/csi/controllerserver.py
@@ -131,6 +131,21 @@ class ControllerServer(csi_pb2_grpc.ControllerServicer):
         pvtype = PV_TYPE_SUBVOL
         is_block = False
 
+        # Debug request.parameters
+        logging.info(logf(
+            "Request Parameters",
+            parameters=request.parameters
+        ))
+        storage_options = None
+        if request.parameters.get("storage_options"):
+            storage_options = request.parameters.get("storage_options")
+            logging.info(logf(
+                "Received storage options",
+                volume=request.name,
+                storage_options=storage_options,
+                type=type(storage_options)
+            ))
+
         # Mounted BlockVolume is requested via Storage Class.
         # GlusterFS File Volume may not be useful for some workloads
         # they can request for the Virtual Block formated and mounted
@@ -386,6 +401,25 @@ class ControllerServer(csi_pb2_grpc.ControllerServicer):
         update_free_size(hostvol, request.name, -pvsize)
 
         send_analytics_tracker("pvc-%s" % hostvoltype, uid)
+
+        # CreateVolumeResponse does not allow to pass
+        # positional arguments or "None"
+        if storage_options:
+            return csi_pb2.CreateVolumeResponse(
+                volume={
+                    "volume_id": request.name,
+                    "capacity_bytes": pvsize,
+                    "volume_context": {
+                        "type": hostvoltype,
+                        "hostvol": hostvol,
+                        "pvtype": pvtype,
+                        "path": vol.volpath,
+                        "fstype": "xfs",
+                        "kformat": kformat,
+                        "storage_options": storage_options
+                    }
+                })
+
         return csi_pb2.CreateVolumeResponse(
             volume={
                 "volume_id": request.name,
@@ -396,10 +430,10 @@ class ControllerServer(csi_pb2_grpc.ControllerServicer):
                     "pvtype": pvtype,
                     "path": vol.volpath,
                     "fstype": "xfs",
-                    "kformat": kformat,
+                    "kformat": kformat
                 }
-            }
-        )
+            })
+
 
     def DeleteVolume(self, request, context):
         start_time = time.time()
--- a/csi/controllerserver.py
+++ b/csi/controllerserver.py
@@ -131,20 +131,7 @@ class ControllerServer(csi_pb2_grpc.ControllerServicer):
         pvtype = PV_TYPE_SUBVOL
         is_block = False
 
-        # Debug request.parameters
-        logging.info(logf(
-            "Request Parameters",
-            parameters=request.parameters
-        ))
-        storage_options = None
-        if request.parameters.get("storage_options"):
-            storage_options = request.parameters.get("storage_options")
-            logging.info(logf(
-                "Received storage options",
-                volume=request.name,
-                storage_options=storage_options,
-                type=type(storage_options)
-            ))
+        storage_options = request.parameters.get("storage_options", "")
 
         # Mounted BlockVolume is requested via Storage Class.
         # GlusterFS File Volume may not be useful for some workloads
@@ -402,24 +389,6 @@ class ControllerServer(csi_pb2_grpc.ControllerServicer):
 
         send_analytics_tracker("pvc-%s" % hostvoltype, uid)
 
-        # CreateVolumeResponse does not allow to pass
-        # positional arguments or "None"
-        if storage_options:
-            return csi_pb2.CreateVolumeResponse(
-                volume={
-                    "volume_id": request.name,
-                    "capacity_bytes": pvsize,
-                    "volume_context": {
-                        "type": hostvoltype,
-                        "hostvol": hostvol,
-                        "pvtype": pvtype,
-                        "path": vol.volpath,
-                        "fstype": "xfs",
-                        "kformat": kformat,
-                        "storage_options": storage_options
-                    }
-                })
-
         return csi_pb2.CreateVolumeResponse(
             volume={
                 "volume_id": request.name,
@@ -430,7 +399,8 @@ class ControllerServer(csi_pb2_grpc.ControllerServicer):
                     "pvtype": pvtype,
                     "path": vol.volpath,
                     "fstype": "xfs",
-                    "kformat": kformat
+                    "kformat": kformat,
+                    "storage_options": storage_options
                 }
             })
 
--- a/csi/nodeserver.py
+++ b/csi/nodeserver.py
@@ -4,7 +4,6 @@ nodeserver implementation
 import logging
 import os
 import time
-import json
 import csi_pb2
 import csi_pb2_grpc
 import grpc
@@ -18,6 +17,8 @@ GLUSTERFS_CMD = "/opt/sbin/glusterfs"
 MOUNT_CMD = "/bin/mount"
 UNMOUNT_CMD = "/bin/umount"
 
+# noqa # pylint: disable=too-many-locals
+# noqa # pylint: disable=too-many-statements
 
 class NodeServer(csi_pb2_grpc.NodeServicer):
     """
@@ -64,14 +65,12 @@ class NodeServer(csi_pb2_grpc.NodeServicer):
         options = request.volume_context.get("options", None)
 
         # Storage volfile options
-        storage_options = request.volume_context.get("storage_options", None)
-
+        storage_options = request.volume_context.get("storage_options", "")
         mntdir = os.path.join(HOSTVOL_MOUNTDIR, hostvol)
 
         pvpath_full = os.path.join(mntdir, pvpath)
 
-
-        logging.info(logf(
+        logging.debug(logf(
             "Received a valid mount request",
             request=request,
             voltype=voltype,
@@ -79,8 +78,7 @@ class NodeServer(csi_pb2_grpc.NodeServicer):
             pvpath=pvpath,
             pvtype=pvtype,
             pvpath_full=pvpath_full,
-            storage_options=storage_options,
-            type=type(storage_options)
+            storage_options=storage_options
         ))
 
         if voltype == "External":
@@ -110,14 +108,8 @@ class NodeServer(csi_pb2_grpc.NodeServicer):
 
         # When 'storage_options' is configured mountpoint & volfile path change,
         # Update pvpath_full accordingly.
-        # debug
-        if storage_options:
+        if storage_options != "":
             pvpath_full = os.path.join(mountpoint, pvpath)
-            logging.debug(logf(
-                "update pvpath_full",
-                pvpath_full=pvpath_full,
-                request_target_path=request.target_path
-            ))
 
         logging.debug(logf(
             "Mounted Hosting Volume",
--- a/csi/nodeserver.py
+++ b/csi/nodeserver.py
@@ -65,9 +65,6 @@ class NodeServer(csi_pb2_grpc.NodeServicer):
 
         # Storage volfile options
         storage_options = request.volume_context.get("storage_options", None)
-        # Convert 'str' to 'dict'
-        if storage_options:
-            storage_options = json.loads(storage_options)
 
         mntdir = os.path.join(HOSTVOL_MOUNTDIR, hostvol)
 
--- a/csi/nodeserver.py
+++ b/csi/nodeserver.py
@@ -4,7 +4,7 @@ nodeserver implementation
 import logging
 import os
 import time
-
+import json
 import csi_pb2
 import csi_pb2_grpc
 import grpc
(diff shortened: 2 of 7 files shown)


csi/volumeutils.py
```diff
@@ -1097,10 +1228,15 @@ def mount_glusterfs(volume, mountpoint, is_client=False):
             "-l", log_file,
             "--volfile-id", volname,
             "--fs-display-name", "kadalu:%s" % volname,
-            "-f", "%s/%s.client.vol" % (VOLFILES_DIR, volname),
+            "-f", client_volfile_path,
```
vatsa287, line 1239: Storage options is being configured and changes are reflected in client volfile for the first request coming in.
After a custom volfile is request is completed, if a default PVC is created with with default SC, error is thrown stating "no such path", since default client volfile is not generated at Line 1121 because of the reason in Line 897.
Can we keep a copy of default or custom client volfile in /tmp or similar?

csi/volumeutils.py
```diff
@@ -1009,13 +1014,140 @@ def reload_glusterfs(volume):
...
         volname = volume["name"]
 
+
+    generate_client_volfile(volname)
+    client_volfile_path = os.path.join(
+        VOLFILES_DIR,
+        "%s.client.vol" % volname
+    )
+    if storage_options is not None:
+
```
aravindavk, line 1138: Use temp file as base file so that original Volfile(with only default options) will remain untouched.

```
import shutil
import tempfile

tmp_volfile = tempfile.mktemp(suffix=".vol")
shutil.copyfile(client_volfile_path, tmp_volfile)

volfile = Volfile.parse(tmp_volfile)
```

Then after finding the hash,

```
os.rename(tmp_volfile, new_client_volfile_path)
```

csi/volumeutils.py
leelavg: - Didn't review yet, just skimming through issues.
- I feel tmpfile couldn't work for container reboot, is there any possibility to use create tmpfile in empty dir
- But above couldn't work for pod reboot though 😅
- Will get back after reviewing the pr

aravindavk: > I feel tmpfile couldn't work for container reboot, is there any possibility to use create tmpfile in empty dir

Persistence not required, only used as base file that gets renamed to actual required file.

> But above couldn't work for pod reboot though

Persistence not required for the generated volfiles, only templates are part of container image.

csi/volumeutils.py
vatsa287:
```bash
root@kadalu-csi-nodeplugin-wxp9c:/kadalu/volfiles# ls
storage-pool-1.client.vol  storage-pool-1_d2435480caff1b10.client.vol
root@kadalu-csi-nodeplugin-wxp9c:/kadalu/volfiles# diff storage-pool-1.client.vol storage-pool-1_d2435480caff1b10.client.vol -wB
32c27
<     option data-self-heal on
---
>     option data-self-heal off
36c31
<     option entry-self-heal on
---
>     option entry-self-heal off
71c64
<     option open-behind on
---
>     option open-behind off
```

csi/controllerserver.py
```diff
@@ -131,6 +131,21 @@ def CreateVolume(self, request, context):
...
         is_block = False
 
+        # Debug request.parameters
+        logging.info(logf(
+            "Request Parameters",
+            parameters=request.parameters
+        ))
+        storage_options = None
+        if request.parameters.get("storage_options"):
+            storage_options = request.parameters.get("storage_options")
```
vatsa287: Storage options are given in the form of map in string, as map is not allowed. Is this OK?
```yaml
kind: StorageClass
apiVersion: storage.k8s.io/v1
metadata:
  name: kadalu.storage-pool-1-vol-options
provisioner: kadalu
parameters:
  storage_options:
    '{
      "cluster/replicate":
      {
        "data-self-heal": "off", "entry-self-heal": "off"
      },
      "performance/open-behind":
      {
        "open-behind": "off"
      }
    }'
```

leelavg: - I couldn't say it's appealing or easy
- How about something like below, not mandatory to be ack'd though 
``` yaml
vol_ops:
  - cluster:
      - replicate:
          data-self-heal: off
          entry-self-heal: off
  - performance:
      - open-behind:
          open-behind: off
```
- Issue might be `off` will be interpreted as `true`

aravindavk: We should use flat structure and Gluster users are familiar with that option names. For some options we may need to add support for aliases.

```yaml
storage_options:
  replicate.data-self-heal: off
  performance.open-behind: off
```

Split the Option name by `.` and then construct the dict when required.

aravindavk: > as map is not allowed

Where is this not allowed? In Storage class or while adding to Volume Context?

vatsa287: strorage class does not allow. 
```bash
error: error validating "vol-options-sc.yaml": error validating data: ValidationError(StorageClass.parameters.storage_options): invalid type for io.k8s.api.storage.v1.StorageClass.parameters: got "map", expected "string"; if you choose to ignore these errors, turn validation off with --validate=false
```
Similarly it does not allow array as well.

csi/volumeutils.py
```diff
@@ -1009,13 +1015,224 @@ def reload_glusterfs(volume):
...
+                    # #option[key] = value
+                    # #print("values, values[key]", values, values[key], option)
+                    # element.options.update(values)
+                    # print("after update", element.name, element.type)
+                    # print(element.options)
+                    # print("\n")
+                    # opt[element.type] = element.options
+                    # # #values = {}, ele.options = {}
+
+                    for key, value in values.items():
```
vatsa287: This check was required to update the option in the element if it only exists, else would add to all matched elements.

csi/volumeutils.py
```diff
@@ -1009,13 +1015,224 @@ def reload_glusterfs(volume):
...
+                    "opts[element.type]",
+                    opts_element_type=opts[element.type],
+                    element_type=element.type
+                ))
+
+    def update_options_by_type2(self, opts):
+        opt = {}
+        for element in self.elements:
+
+            # Example: 'cluster/replicate' & 'performance/open-behind'
```
aravindavk: You can't hard code performance here. There are many xlators where option name is same as xlator name to switch off/on. For example: `features/changelog.changelog=on`

For now accept Volume options as `performance/open-behind.open-behind: on`. Introduce alias support then we can have a json mapping for Option name and expanded option names. 

All supported options(Also helps us to validate)

```json
{
    "performance.open-behind": "performance/open-behind.open-behind"
}
```

aravindavk: Alias support can be in next PR. This PR can accept fully expanded Option names only.

vatsa287:
@amarts , @aravindavk , @leelavg ^^

csi/volumeutils.py
vatsa287: done

csi/controllerserver.py
```diff
@@ -131,6 +131,10 @@ def CreateVolume(self, request, context):
         pvtype = PV_TYPE_SUBVOL
         is_block = False
 
+        storage_options = None
```
aravindavk: Replace these three lines with

```
storage_options = request.parameters.get("storage_options", "")
```

csi/controllerserver.py
```diff
@@ -386,6 +390,25 @@ def CreateVolume(self, request, context):
         update_free_size(hostvol, request.name, -pvsize)
 
         send_analytics_tracker("pvc-%s" % hostvoltype, uid)
+
+        # CreateVolumeResponse does not allow to pass
+        # positional arguments or "None"
+        if storage_options:
```
aravindavk: You can remove this now.

csi/nodeserver.py
```diff
@@ -63,6 +62,9 @@ def NodePublishVolume(self, request, context):
         gvolname = request.volume_context.get("gvolname", None)
         options = request.volume_context.get("options", None)
 
+        # Storage volfile options
+        storage_options = request.volume_context.get("storage_options", None)
```
aravindavk: Use `""` instead of `None`

csi/nodeserver.py
```diff
@@ -98,7 +102,19 @@ def NodePublishVolume(self, request, context):
...
         }
 
-        mount_glusterfs(volume, mntdir, True)
+
+        mountpoint = mount_glusterfs(volume, mntdir, storage_options, True)
+
+        # When 'storage_options' is configured mountpoint & volfile path change,
+        # Update pvpath_full accordingly.
+        # debug
+        if storage_options:
```
aravindavk: `if storage_options != "":`

csi/volumeutils.py
```diff
@@ -1009,13 +1015,156 @@ def reload_glusterfs(volume):
...
+
+    def update_options_by_type(self, opts):
+        """
+        Update existing storage-options with the options
+        passed by user through storage-class
+        """
+
+        opt = {}
+        for element in self.elements:
+            if opts.get(element.type, None) is not None:
```
aravindavk, line 1066: Replace below lines with

```
element.options.update(opts.get(element.type, {}))
```

csi/volumeutils.py
```diff
@@ -1009,13 +1015,156 @@ def reload_glusterfs(volume):
...
         volname = volume["name"]
 
+
+    generate_client_volfile(volname)
+    client_volfile_path = os.path.join(
+        VOLFILES_DIR,
+        "%s.client.vol" % volname
+    )
+
+    if storage_options is not None:
```
aravindavk: `if storage_options != ""`

csi/volumeutils.py
```diff
@@ -1009,13 +1015,156 @@ def reload_glusterfs(volume):
...
+        tmp_volfile_path = tempfile.mkstemp()[1]
+        shutil.copy(client_volfile_path, tmp_volfile_path)
+
+        # Parse the client-volfile, update passed storage-options & save
+        parsed_client_volfile_path = Volfile.parse(tmp_volfile_path)
+        parsed_client_volfile_path.update_options_by_type(storage_options)
+        parsed_client_volfile_path.save()
+
+        # Sort storage-options and generate hash
+        storage_options_sorted = dict(sorted(storage_options.items()))
```
aravindavk: `json.dumps` accepts `sort_keys=True` option

csi/volumeutils.py
vatsa287: Ack to rest all suggestions.
Here if we update without checking if the key is present, the options will be added to all matched types.
Example:
If `protocol/client.remote-port: NNNN` or `protocol/client.volfile-key: XYZ` is given where `protocol/client` has many instances, these changes will be reflected in all those instances. So check for given key exist is important in my opinion.

```yaml
volume 09d204a6-4ba4-11ec-ac5e-0242ac110003-ta
    type protocol/client
    option transport.socket.read-fail-log false
    option remote-host tie-breaker.kadalu.io
    option remote-port 24007
    option volfile-key /storage-pool-1
    option remote-subvolume /mnt
end-volume
volume storage-pool-1-client-0
    type protocol/client
    option transport.socket.read-fail-log false
    option volfile-key /storage-pool-1
    option remote-subvolume /bricks/storage-pool-1/data/brick
    option remote-host server-storage-pool-1-0-0.storage-pool-1
end-volume
volume storage-pool-1-client-1
    type protocol/client
    option transport.socket.read-fail-log false
    option volfile-key /storage-pool-1
    option remote-subvolume /bricks/storage-pool-1/data/brick
    option remote-host server-storage-pool-1-1-0.storage-pool-1
end-volume
```
Here `volfile-key` at 1st element superfluous.

aravindavk: What are the options provided by the user? 

It looks correct to me(Also I am not able to find the difference between oneline version and expanded one, expanded code does not add a option if it is already exists?)

Some of the options are not for customization. For example, port, path, volfile-key etc. If name level customization is required then either use update by name or Add `glusterfs` option `--xlator-option=*-client-0.volfile-key=XYZ`

vatsa287: The first element `volume 09d204a6-4ba4-11ec-ac5e-0242ac110003-ta` did not have the option `volfile-key`. Yes just using this as an example. Works fine for elements are not repeated like here `protocol/client` is present 3 times, rest are only one.

amarts (commented):
can we add one case of testing with this option?

also, need to test it with more nodes, like what happens if we have 9 storage pods, with replica 3 and see if all are applied.

amarts (approved):
LGTM

csi/volumeutils.py
```diff
@@ -1009,21 +1011,166 @@ def reload_glusterfs(volume):
...
+        for element in self.elements:
+            if opts.get(element.type, None) is not None:
+                element.options.update(opts.get(element.type, {}))
+
+                # for key, value in opts[element.type].items():
+                #     if element.options.get(key):
+                #         option = {}
+                #         option[key] = value
+                #         element.options.update(option)
+                #         opt[element.type] = element.options
```
leelavg: - Pls replace with TODO if there's anything missing or else remove the comments

csi/volumeutils.py
```diff
@@ -1009,21 +1011,166 @@ def reload_glusterfs(volume):
...
+                if line.startswith("volume "):
+                    element = VolfileElement(line.split()[-1])
+                elif line == "end-volume":
+                    elements.append(element)
+                    element = None
+                elif line.startswith("option "):
+                    _, name, value = line.split()
+                    element.options[name] = value
+                elif line.startswith("subvolumes "):
+                    element.subvolumes = line.split(" ", 1)[-1]
```
leelavg, line 1053: - I might be wrong, does `subvolumes` always contain a single entry?
- I faintly remember there can be many clients in a subvolumes 🤔 ?

leelavg (commented):
- only minor comments, lgtm.
- in general, do you see any side-effect for the current mounts when kadalu is upgraded?

csi/volumeutils.py
vatsa287: Yes it contains multiple values. I don't think this is comes under supported volume options. Anyway @aravindavk has to confirm the list of options that will be supported by kadalu.

csi/volumeutils.py
vatsa287: Ack, Kept this if it is necessary to check if the option exist by default to update or not.

csi/volumeutils.py
aravindavk: it is a flat structure. Each volume entry in volfile will have only one line for subvolumes and each subvolume(Volfile subvolume) are separated by space.

Example: subvolumes replicate-spool1-0 replicate-spool1-1

leelavg: - then why are we storing only the last entry or I missed something in the parsing logic?

aravindavk: Split(" ", 1) is used, all subvol list is kept unchanged and added back later.

leelavg (approved):
/lgtm

aravindavk (approved):

PR #715, opened 2021-11-28
Fix kadalu-info ConfigMap on-update reload

### LINKED ISSUE #714: Support for Replica1 to Replica3 migration
Hello all!
I've experimented a little bit with Kadalu and everything seems really good moreover the performances are great compared to other CAS/CSI supporting RWX.
During my experimentations I've found an issue, maybe it is related to my specific use case, I know a thing or two about GlusterFS and I've done volume replica addition in the past.
To reproduce the issue:
- Create a single node cluster and install Kadalu
- Create a KadaluStorage resource with the following manifest:
``` yaml
kind: KadaluStorage
metadata:
  name: storage01
  namespace: kadalu
spec:
  type: Replica1
  pvReclaimPolicy: archive
  storage:
    - node: node1
      device: /dev/kadalu/kadalu
```
- Create PVCs, Test Kadalu, everything ok
- Join some more nodes to the cluster
- Apply/Patch KadaluStorage with the following manifest in order to achieve Replication across nodes:
``` yaml
kind: KadaluStorage
metadata:
  name: storage01
  namespace: kadalu
spec:
  type: Replica3
  pvReclaimPolicy: archive
  storage:
    - node: node1
      device: /dev/kadalu/kadalu
    - node: node2
      device: /dev/kadalu/kadalu
    - node: node3
      device: /dev/kadalu/kadalu
```
- The KadaluStorage shows Replica3
- `kubectl kadalu healinfo` says: `This command is supported for only volumes of replicate/disperse type. Volume storage01 is not of type replicate/disperse`

Let me know if I can do something more to help you understand my issue!

MatteoManzoni:
Following the discussion in the related issue (#714) this PR proposes one of the possible solutions to address kadalu-info on-update reload.
The proposed solution is based on establishing the inode watchers on `/var/lib/gluster/..data` instead of `/var/lib/gluster`, another possible solution is to establish said watchers recursively on `/var/lib/gluster`

Tested on:
- k8s v1.20.12
- k3s v1.22.3+k3s1

Fixes: #714

### DIFF
--- a/csi/watch-vol-changes.sh
+++ b/csi/watch-vol-changes.sh
@@ -6,19 +6,19 @@
 # the pod.
 
 # ConfigMap dir as per csi.yaml pod's template.
-CONFIG_MAP_DIR="/var/lib/gluster"
+CONFIG_MAP_DIR="/var/lib/gluster/..data"
 
 # Currently 'SIGHUP' handling is done only in 'csi/main.py'.
-CSI_PROCESS_ID=$(ps aux|grep python | grep main | awk '{ print $2}' | xargs);
+CSI_PROCESS_ID="$(pgrep -f '^python3.*main.py$')";
 
 echo "Starting watch on configmap"
 while true; do
     # Starting inotifywait without `-m` option (ie, `--monitor`), which makes the process
     # exit after first instance of modify on directory.
-    line=$(inotifywait -e modify ${CONFIG_MAP_DIR});
+    line="$(inotifywait -e modify ${CONFIG_MAP_DIR})";
     # Send a blanket HUP, so all vols can be checked and relevant glusterfs process can be sent a SIGHUP after volgen
-    echo $line
-    kill -HUP $CSI_PROESS_ID;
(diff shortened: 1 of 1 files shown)


leelavg (approved):

leelavg:
@MatteoManzoni thanks for your contribution 😃 

- going by the reasoning mentioned in https://github.com/kadalu/kadalu/issues/714#issuecomment-981147133 it's appropriate to watch on `data` dir based on
1. unless there's a change to how config map work internally from k8s side this'll work
2. recursive is only a last option and should be discouraged as such 😅

amarts (approved):

vatsa287 (approved):

aravindavk (approved):

denogio:
This PR breaks kadalu from working in a Nomad environment. There is no file or directory named ..data in a nomad environment. I would propose to setup a recursive watch of /var/lib/gluster instead, and put a sleep 1 into the loop, because else it has the potential of becoming very CPU intensive.

PR #717, opened 2021-11-30
Additional check in ControllerExpandVolume

rafis:
```
[2021-11-30 12:35:03,954] ERROR [main - 42:mount_storage] - Unable to mount volume       hvol=out-dev-gluster
[2021-11-30 12:35:03,962] INFO [main - 76:main] - Server started
[2021-11-30 12:35:04,787] ERROR [_server - 454:_call_behavior] - Exception calling application: 'NoneType' object has no attribute 'extra'
  File "/kadalu/lib/python3.9/site-packages/grpc/_server.py", line 444, in _call_behavior
  File "/kadalu/controllerserver.py", line 601, in ControllerExpandVolume
    if existing_volume.extra['kformat'] == 'non-native':
```

### DIFF
--- a/csi/controllerserver.py
+++ b/csi/controllerserver.py
@@ -598,6 +598,16 @@ class ControllerServer(csi_pb2_grpc.ControllerServicer):
 
         # Get existing volume
         existing_volume = search_volume(request.volume_id)
+        if not existing_volume:
+            errmsg = logf(
+                "Unable to find volume",
+                volume_id=request.volume_id
+            )
+            logging.error(errmsg)
+            context.set_details(str(errmsg))
+            context.set_code(grpc.StatusCode.INVALID_ARGUMENT)
+            return csi_pb2.ControllerExpandVolumeResponse()
+
         if existing_volume.extra['kformat'] == 'non-native':
             errmsg = "PV with kadalu_format == non-native doesn't support Expansion"
             logging.error(errmsg)


amarts (approved):
Looks like a `-s` with commit is required to pass the checks.

leelavg:
@rafis
- just reasoning through how this is hit 🤔 
- can you pls let me know how this has happened?
- it seems current pr isn't the root cause (not against having this check though)
- error is in the first line, why kadalu isn't able to mount the volume :/

rafis:
Because I don't know what is the value of `volume_id`, I don't understand where it is coming from. Is it from `kubectl -n kadalu get configmap kadalu-info -o yaml`?

```
apiVersion: v1
data:
  out-dev-gluster.info: '{"volname": "out-dev-gluster", "volume_id": "80a1d68a-1f74-11ec-a43a-b644e6b66ae1",
    "type": "External", "pvReclaimPolicy": "delete", "kadalu_format": "native", "gluster_hosts":
    "10.10.10.12,10.10.10.13,10.10.10.11", "gluster_volname": "kadalu"}'
  uid: 81ff8963-87a0-4a58-bd9e-0e9cb98bcd95
  volumes: ""
kind: ConfigMap
metadata:
  creationTimestamp: "2021-09-27T09:22:14Z"
  name: kadalu-info
  namespace: kadalu
  resourceVersion: "74343015"
  uid: 0dc7f289-3153-4eb4-8d19-4f1f9409af0f
```

It was configure by previous SysOps engineer, so it hard for me to understand. Is it volume_id of gluster volume `kadalu`?

leelavg:
> Because I don't know what is the value of volume_id,

- that's not the issue here, explaining a bit more of what's happened

- provisioner pod restarted but volumes created before the restart were not able to mount in the restarted provisioner
- after restart k8s asked for a pvc expansion but as the underlying volume isn't mounted you hit above error
- connecting the dots here, this might be related to #719 and thanks for the extra check as the thing we don't want is an Exception triggering container restarts 😅

leelavg (approved):

PR #736, opened 2022-01-06
Fix kadalu update issue

### LINKED ISSUE #735: Cannot update kaDalu because changed immutable parameter on CSIDriver resource
Operator is in loop with this error:

2022-01-04 18:46:37,249] INFO [main - 958:deploy_config_map] - ConfigMap Deployed	 manifest=/kadalu/templates/configmap.yaml uid=274bff60-41f8-4dba-aa51-268a6bb7e1a4 upgrade=True
  File "/kadalu/main.py", line 1029, in <module>
    main()
  File "/kadalu/main.py", line 1005, in main
    deploy_csi_pods(core_v1_client)
  File "/kadalu/main.py", line 904, in deploy_csi_pods
    lib_execute(KUBECTL_CMD, APPLY_CMD, "-f", filename)
  File "/kadalu/kadalulib.py", line 156, in execute
kadalulib.CommandException: [1]  The CSIDriver "kadalu" is invalid: spec.podInfoOnMount: Invalid value: true: field is immutable
[2022-01-04 18:46:38,449] INFO [kadalulib - 405:monitor_proc] - Restarted Process	 name=operator
[2022-01-04 18:46:39,176] INFO [main - 936:deploy_config_map] - Found existing configmap. Updating	 name=kadalu-info
[2022-01-04 18:46:39,201] INFO [main - 958:deploy_config_map] - ConfigMap Deployed	 manifest=/kadalu/templates/configmap.yaml uid=274bff60-41f8-4dba-aa51-268a6bb7e1a4 upgrade=True

I'm not sure if I can delete CSIDriver object without losing all my PVC's. The field podInfoOnMount is immutable.

Running kaDalu 0.8.9 but same issue with 0.8.7 or 0.8.8

vatsa287:
Fixes: #735

### DIFF
--- a/operator/main.py
+++ b/operator/main.py
@@ -838,6 +838,27 @@ def delete_storage_class(hostvol_name, _):
     ))
 
 
+def csi_driver_object_api_version():
+    """
+    Return API Version of CSI Driver object"
+    """
+
+    cmd = ["kubectl", "get", "csidriver", "kadalu", "-ojson"]
+
+    try:
+        resp = utils_execute(cmd)
+        csi_driver_data = json.loads(resp.stdout)
+        version = csi_driver_data["apiVersion"]
+        return version
+
+    except CommandError as err:
+        logging.error(logf(
+            "Failed to get version of csi driver object",
+            error=err
+        ))
+        return None
+
+
 def watch_stream(core_v1_client, k8s_client):
     """
     Watches kubernetes event stream for kadalustorages in Kadalu namespace
@@ -899,6 +920,11 @@ def deploy_csi_pods(core_v1_client):
 
     if api_instance.major > "1" or api_instance.major == "1" and \
        api_instance.minor >= "22":
+
+        if csi_driver_object_api_version() is not None and \
+           csi_driver_object_api_version() != "v1":
+            lib_execute(KUBECTL_CMD, DELETE_CMD, "csidriver", "kadalu")
+
         filename = os.path.join(MANIFESTS_DIR, "csi-driver-object-v1.yaml")
         template(filename, namespace=NAMESPACE, kadalu_version=VERSION)
         lib_execute(KUBECTL_CMD, APPLY_CMD, "-f", filename)
--- a/operator/main.py
+++ b/operator/main.py
@@ -921,9 +921,14 @@ def deploy_csi_pods(core_v1_client):
     if api_instance.major > "1" or api_instance.major == "1" and \
        api_instance.minor >= "22":
 
-        if csi_driver_object_api_version() is not None and \
-           csi_driver_object_api_version() != "v1":
+        csi_driver_version = csi_driver_object_api_version()
+        if csi_driver_version is not None and \
+           csi_driver_version != "v1":
             lib_execute(KUBECTL_CMD, DELETE_CMD, "csidriver", "kadalu")
+            logging.info(logf(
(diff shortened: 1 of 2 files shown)


aravindavk (approved):

templates/csi-driver-object.yaml.j2
```diff
@@ -3,5 +3,5 @@ kind: CSIDriver
 metadata:
   name: kadalu
 spec:
-  attachRequired: true
+  attachRequired: false
```
leelavg: don't change this or else existing deployments on operator restart will fail for k8s 1.14 to 1.21

operator/main.py
```diff
@@ -899,6 +919,12 @@ def deploy_csi_pods(core_v1_client):
 
     if api_instance.major > "1" or api_instance.major == "1" and \
        api_instance.minor >= "22":
+
+        if csi_driver_object_api_version() is not None and \
+           csi_driver_object_api_version() == "v1beta1":
```
leelavg: better check for our intention, like we want `v1` in this k8s range, so it'll be `!= "v1"`

operator/main.py
```diff
@@ -837,6 +837,26 @@ def delete_storage_class(hostvol_name, _):
         volname=hostvol_name
     ))
 
+def csi_driver_object_api_version():
+    """
+    Return API Version of CSI Driver object"
+    """
+
+    cmd = ["kubectl", "get", "csidriver", "-nkadalu", "-ojson"]
```
leelavg: - incorrect command, csidriver is a cluster scoped resource and more over search for kadalu
- if it's a typo just change `-nkadalu` => `kadalu`

leelavg (changes requested):

templates/csi-driver-object.yaml.j2
vatsa287: Ack

operator/main.py
vatsa287: Ack.

operator/main.py
vatsa287: Yeah maybe because it can be anything other than `v1beta1` as well. Changed it.

operator/main.py
```diff
@@ -899,6 +920,12 @@ def deploy_csi_pods(core_v1_client):
 
     if api_instance.major > "1" or api_instance.major == "1" and \
        api_instance.minor >= "22":
+
+        if csi_driver_object_api_version() is not None and \
+           csi_driver_object_api_version() != "v1":
+            filename = os.path.join(MANIFESTS_DIR, "csi-driver-object.yaml")
```
leelavg: - we know the csi driver name is `kadalu` no need to use file
- just perform `k delete csidriver kadalu`
- current condition will fail if we are updating from <=1.14

Pls don't mind, being extra cautious after gluster mount options debacle 😅

leelavg (changes requested):

operator/main.py
vatsa287: Sure ack.

vatsa287: lib_execute(*cmd) works or do you recommend to use non file use based commands on utis_execute()?

leelavg: fine with anything that works and there's no need for a function for that single command 😅

leelavg (approved):

PR #745, opened 2022-01-14
Multiple fixes for External mount

fastlorenzo:
1. Fixed non-existing "bricks" and "decommissioned" keys when using External glusterfs storage.

    Example of content for `data` with an external storage (`root@kadalu-csi-provisioner-0:/# cat /var/lib/gluster/data.info`: 
    ```json
    {"volname": "data", "volume_id": "f4b37c38-7517-11ec-b01e-f66bd7a0c157", "type": "External", "pvReclaimPolicy": "delete", "kadalu_format": "non-native", "gluster_hosts": "172.17.0.1", "gluster_volname": "data", "gluster_options": "log-level=DEBUG"}
    ```

2. Fixed duplicate mount path in the command line parameters

### DIFF
--- a/csi/volumeutils.py
+++ b/csi/volumeutils.py
@@ -1294,8 +1294,6 @@ def mount_glusterfs_with_host(volname, mountpoint, hosts, options=None, is_clien
         for option in options.split(","):
             g_ops.append(f"--{option}")
 
-    cmd.append(mountpoint)
-
     logging.debug(logf(
         "glusterfs command",
         cmd=cmd,
--- a/csi/volumeutils.py
+++ b/csi/volumeutils.py
@@ -1124,6 +1124,54 @@ def mount_glusterfs(volume, mountpoint, storage_options="", is_client=False):
     else:
         volname = volume["name"]
 
+    if volume['type'] == 'External':
+        # Try to mount the Host Volume, handle failure if
+        # already mounted
+        with mount_lock:
+            mount_glusterfs_with_host(volname,
+                                      mountpoint,
+                                      volume['g_host'],
+                                      volume['g_options'],
+                                      is_client)
+        use_gluster_quota = False
+        if (os.path.isfile("/etc/secret-volume/ssh-privatekey")
+            and "SECRET_GLUSTERQUOTA_SSH_USERNAME" in os.environ):
+            use_gluster_quota = True
+        secret_private_key = "/etc/secret-volume/ssh-privatekey"
+        secret_username = os.environ.get('SECRET_GLUSTERQUOTA_SSH_USERNAME', None)
+
+        # SSH into only first reachable host in volume['g_host'] entry
+        g_host = reachable_host(volume['g_host'])
+
+        if g_host is None:
+            logging.error(logf("All hosts are not reachable"))
+            return
+
+        if use_gluster_quota is False:
+            logging.debug(logf("Do not set quota-deem-statfs"))
+        else:
+            logging.debug(logf("Set quota-deem-statfs for gluster directory Quota"))
+            quota_deem_cmd = [
+                "ssh",
+                "-oStrictHostKeyChecking=no",
+                "-i",
+                "%s" % secret_private_key,
+                "%s@%s" % (secret_username, g_host),
+                "sudo",
+                "gluster",
+                "volume",
+                "set",
+                "%s" % volume['g_volname'],
+                "quota-deem-statfs",
+                "on"
+            ]
+            try:
+                execute(*quota_deem_cmd)
+            except CommandException as err:
+                errmsg = "Unable to set quota-deem-statfs via ssh"
+                logging.error(logf(errmsg, error=err))
+                raise err
+        return mountpoint
 
     generate_client_volfile(volname)
     client_volfile_path = os.path.join(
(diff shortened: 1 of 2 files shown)


fastlorenzo:
Linking to #562

csi/volumeutils.py
```diff
@@ -1294,8 +1294,6 @@ def mount_glusterfs_with_host(volname, mountpoint, hosts, options=None, is_clien
         for option in options.split(","):
             g_ops.append(f"--{option}")
 
-    cmd.append(mountpoint)
```
leelavg, line 1297: - I don't see this as a duplicate, can you please expand on why it's being removed?

leelavg (commented):
@fastlorenzo welcome to kadalu and thanks for the contribution!

csi/volumeutils.py
leelavg: Now I see it, you are correct, let's see why CI is failing

fastlorenzo: great, for reference it was this line: `command = cmd + g_ops + [mountpoint]` already adding it

fastlorenzo:
Thanks, and thank you for this great tool 😄 
I hope I can contribute to make it work with k3s and the use case I have: kubernetes cluster deployed with k3d (several k3s nodes running on docker) with a glusterfs share on the host system, accessed by the nodes.

leelavg:
> kubernetes cluster deployed with k3d
- good to know fellow k3d user and I use k3d for testing & development of kadalu, you can see the write up [here](https://thoughtexpo.com/tags/kubernetes/) if you are interested in knowing the internals (posts are quickly becoming outdated with kadalu releases 😅)
- I assume you have a reason for using k3d and may I know the reason for such a short lived system?
- Btw, I hope you are not on MacOS as k3d needs a shared mount for any CSI to work afair
- Reg CI failure need to look into it when I'm near my workstation, not sure force push fixes it or it's a genuine failure

fastlorenzo:
- Nice write-ups 😄 
- Yes, I'm replacing my old VPS (using Proxmox/OpenVZ) and initially looked at dockerizing the infra. I then decided to move to kubernetes for ease of management and upgrade. I'm familiar with k3s as I have a raspberry Pi cluster at home, running on k3s (using NFS storage for persistence).
- Nope, that should be fine 👍 
- I've had a quick look but tbh I'm a bit lost in the huge output 😉

csi/volumeutils.py
```diff
@@ -1124,6 +1124,54 @@ def mount_glusterfs(volume, mountpoint, storage_options="", is_client=False):
...
+            ]
+            try:
+                execute(*quota_deem_cmd)
+            except CommandException as err:
+                errmsg = "Unable to set quota-deem-statfs via ssh"
+                logging.error(logf(errmsg, error=err))
+                raise err
+        return mountpoint
 
     generate_client_volfile(volname)
```
leelavg, line 1180: - as `mount_glusterfs` is refactored, changes to `generate_client_volfile` will not have any effect now
- so, we shouldn't be tolerating absence of `bricks` in kadalu-info cm and it should fail to generate client volfile

csi/volumeutils.py
```diff
@@ -1124,6 +1124,54 @@ def mount_glusterfs(volume, mountpoint, storage_options="", is_client=False):
     else:
         volname = volume["name"]
 
+    if volume['type'] == 'External':
```
leelavg, line 1131: - IMO moving this check here is the right thing to do
- it calls for some extra changes as well, from existing code perspective there are some places where `mount_glusterfs_with_host` is directly invoked and those will fail due to this refactor
```
-> rg mount_glusterfs_with_host
csi/volumeutils.py
1131:            mount_glusterfs_with_host(volname,
1262:def mount_glusterfs_with_host(volname, mountpoint, hosts, options=None, is_client=False):
1358:    mount_glusterfs_with_host(hvol['g_volname'], mntdir, hvol['g_host'], hvol['g_options'])

csi/nodeserver.py
10:from volumeutils import (mount_glusterfs, mount_glusterfs_with_host,
87:                mount_glusterfs_with_host(gvolname, request.target_path, gserver, options, True)
```
- referring above, pls make changes in `csi/volumeutils.py:1131` and `csi/nodeserver.py:87` to use `mount_glusterfs` rather than the other method, I hope this makes sense.

leelavg (changes requested):

leelavg:
- even though this pr doesn't touch test codepath not sure why CI is failing will re-look at the earliest

csi/volumeutils.py
fastlorenzo: I've checked all calls to `mount_glusterfs_with_host`:
1. `csi/volumeutils.py:1131` => not necessary to change as this is within `mount_glusterfs´ function
2. `csi/nodeserver.py:87`: I'll tackle this one, I just need to check the best way to do it
3. `csi/volumeutils.py:1358`: I guess this one needs to be updated as well

csi/volumeutils.py
fastlorenzo: Maybe an explicit check for the `bricks` should be done then, that would make the code clearer

csi/volumeutils.py
leelavg: yes, I meant 1358 in place of 1131, thanks for checking.

fastlorenzo: Pushed proposed changes, this should simplify the code as well

csi/volumeutils.py
fastlorenzo: Should there be bricks for all types except `External`? If so, I can implement a check in `generate_client_volfile` to handle it

leelavg:
@aravindavk I have some other deadline tasks and so couldn't concentrate on this PR, can you have a look at it?

aravindavk:
> @aravindavk I have some other deadline tasks and so couldn't concentrate on this PR, can you have a look at it?

Ack. Will look into this today. Thanks

leelavg:
- pls rebase, resolved CI issue

fastlorenzo:
> * pls rebase, resolved CI issue

Done

amarts (commented):
LGTM

fastlorenzo:
Added one last fix if PV is created manually and fixed a regression

fastlorenzo (commented):
ok to merge I think

leelavg:
/hold

- another makedirs call isn't needed that'll be done as part of `create_(subdir|block)` 
- referring to the manifest provided in slack, IMO there's no need to provide `path` in pv manifest
- you just need to provide below (or similar for external mounts) and xxhash is idempotent for same name
```
2022-01-17T06:04:23.1039654Z [2022-01-17 06:03:59,684] DEBUG [controllerserver - 100:CreateVolume] - Create Volume request	 request=name: "pvc-fba0358f-f0c6-41a4-baa0-c3de0b31e6ae"
2022-01-17T06:04:23.1041363Z capacity_range {
2022-01-17T06:04:23.1041589Z   required_bytes: 838860800
2022-01-17T06:04:23.1041811Z }
2022-01-17T06:04:23.1042011Z volume_capabilities {
2022-01-17T06:04:23.1042235Z   mount {
2022-01-17T06:04:23.1042414Z   }
2022-01-17T06:04:23.1042617Z   access_mode {
2022-01-17T06:04:23.1042856Z     mode: SINGLE_NODE_WRITER
2022-01-17T06:04:23.1043062Z   }
2022-01-17T06:04:23.1043248Z }
2022-01-17T06:04:23.1043430Z parameters {
2022-01-17T06:04:23.1043658Z   key: "kadalu_format"
2022-01-17T06:04:23.1043868Z   value: "native"
2022-01-17T06:04:23.1044074Z }
2022-01-17T06:04:23.1044256Z parameters {
2022-01-17T06:04:23.1044481Z   key: "storage_name"
2022-01-17T06:04:23.1044825Z   value: "storage-pool-4"
2022-01-17T06:04:23.1045035Z }
```

csi/volumeutils.py
leelavg: 1. Yes, except for External there should be bricks field in CR.
2. For external we delegate pulling volfile from external gluster to gluster binary itself
3. Ack, validating bricks field in `generating_client_volfile` seems sensible and make it return false on failure and stop proceeding with mounting volume/glusterfs.

csi/volumeutils.py
```diff
@@ -824,6 +824,10 @@ def volume_list(voltype=None):
 def mount_volume(pvpath, mountpoint, pvtype, fstype=None):
     """Mount a Volume"""
 
+    # Create subvol dir if PV is manually created
+    if not os.path.exists(pvpath):
```
leelavg, line 828: - I couldn't think through why this is needed here.
- `mount_volume` is called after `create_(subdir|block)_volume` only and so pvpath will be created.

fastlorenzo: I'm not sure why, but without this the folder is never created when I supply the manifest for the PV and PVC

fastlorenzo:
Path is actually needed as it's used by mount_volume if you're using subvol mode

leelavg (approved):
/lgtm
- ack, there are some minor issues which doesn't concern this PR. Good to be merged.

PR #750, opened 2022-01-19
Add 'retain' reclaim policy

### LINKED ISSUE #747: [Enhancement] Add possibility to 'retain' volume on delete
It would be great to have the possibility to use 'retain' mode in addition to 'delete' and 'archive' when a PV is deleted.

After investigating the code, I think this should only require minor changes.

My example use case would be: I'm using k3d to deploy a k3s cluster, with glusterfs installed on the host. when deleting the cluster, I'd like to keep the data in the original folders on glusterfs (subvol).
This would also require another enhancement: possibility to set the path name pattern for the mount paths (subvol) - similar to how we can do with https://github.com/kubernetes-sigs/nfs-subdir-external-provisioner `pathPattern` option. (cf. #748)

fastlorenzo:
Fixes #747 (includes fixes from #745)

### DIFF
--- a/cli/kubectl_kadalu/storage_add.py
+++ b/cli/kubectl_kadalu/storage_add.py
@@ -40,7 +40,7 @@ def set_args(name, subparsers):
         default=None)
     arg("--pv-reclaim-policy",
         help="PV Reclaim Policy",
-        choices=["delete", "archive"],
+        choices=["delete", "archive", "retain"],
         default=None)
     arg("--device",
         help=("Storage device in <node>:<device> format, "
--- a/csi/volumeutils.py
+++ b/csi/volumeutils.py
@@ -652,6 +652,15 @@ def delete_volume(volname):
 
     volpath = os.path.join(HOSTVOL_MOUNTDIR, vol.hostvol, vol.volpath)
 
+    # Stop the delete operation if the reclaim policy is set to "retain"
+    if pv_reclaim_policy == "retain":
+        logging.info(logf(
+            "'retain' reclaim policy, volume not deleted",
+            volpath=volpath,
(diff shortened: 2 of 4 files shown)


leelavg:
- Looked at #745 today as well
- Mostly it just needs a timeout increase, if the CI fails can you increase the value from 400 to 600 in https://github.com/kadalu/kadalu/blob/devel/tests/minikube.sh#L342
- Pardon the inconvenience, totally into something else and context switches not working well 😅

fastlorenzo:
Thanks and done ;)

fastlorenzo:
@leelavg it looks like 600 might not be enough yet, I'll try to increase to 800.
Could it be related to all the DNS resolution issues?
`[2022-01-19 15:27:35.160093 +0000] E [name.c:264:af_inet_client_get_remote_sockaddr] 0-storage-pool-4-client-1: DNS resolution failed on host server-storage-pool-4-1-0.storage-pool-4`

leelavg:
- No, timeout isn't the issue.
~- Taking a closer look again, logs mentions a storage pool is mounted but next line states filenotfound error~
~- So, most probably something going wrong while verifying gluster process is running or not, like `ps` not found on container~

fastlorenzo:
rebased as well for CI fix

amarts (commented):
LGTM

Looks like this includes content from other PR also. Ok with merging, but considering squash merge happens, would be good to check if we want to keep them as separate.

fastlorenzo:
indeed, this one is rebased on my 'devel' branch (the other PR)

fastlorenzo:
Rebased changes with `devel` branch

amarts (approved):

PR #751, opened 2022-01-20
test: refactor and modify test code for readability

leelavg:
- run `shfmt` on bash test file
- fail the test when first failure is occurred
- add labels to pvc and pods deployed during test
- prefer using kubectl wait rather than bash polling
- move test units into individual functions
- do not test storage options temporarily
- make sure `restartPolicy: OnFailure` exists on all test app pods

Signed-off-by: Leela Venkaiah G <leelavg@thoughtexpo.com>

PR #753, opened 2022-01-20
CI: Add sleep after pods are in Running state

amarts:


leelavg:
- thanks for the help with finding cause of CI issue (~storage options) and closing in favor of #751

PR #755, opened 2022-01-25
cli: Add support to remove archived pvcs

### LINKED ISSUE #746: How to remove archived volume?
I have configured storage pool with `pvReclaimPolicy: archive`. Created, then deleted pv, now it is archived. How do I remove archived volume to reclaim used space?

vatsa287:
Fixes: #746


leelavg:
- Overall the approach seems to be too much of an effort
- You can disagree and at the same time below is what I anticipate
1. We'll have a new standalone python script which'll be part of CSI image
2. This'll import required methods from volumeutils and perform removal of archive pvcs, info file and update sqlite3
3. Script can be invoked via CLI with python and receive arguments whether to delete every archived vol in a pool or a specific one
4. You can add filters to yield methods to only return archived PVCs etc

vatsa287:
@leelavg Ack to the reviews. Just gave a try with this method. Came to know that by having the logic inside CSI as a script serves for users with no CLI installed. 
Just one query, For PVC specific deletion, how to get the associated name of PVC to be deleted in its pvc-random-uuid format? AFAIK the yield_ method returns its uuid.

leelavg:
> Just one query, For PVC specific deletion, how to get the associated name of PVC to be deleted in its pvc-random-uuid format? AFAIK the yield_ method returns its uuid.

- that's only for the users who actually know the archived PV holding the data
- let the user supply PV name itself and match it against archived PVC

cli/kubectl_kadalu/remove_archived_pv.py
```diff
@@ -0,0 +1,107 @@
...
+    ]
+
+    print("hello", cmd)
+
+    try:
+        resp = utils.execute(cmd)
+        print(
+            "Sent request for deletion of archived PVCs"
+        )
+        print(resp.stderr, resp.stdout, resp.returncode)
```
vatsa287: @leelavg Wanted to capture details here in order to confirm if actually archived PVCs are found & deleted, handle any errors. From Csi at line no. 18, 20

csi/remove_archived_pv.py
```diff
@@ -0,0 +1,64 @@
...
+        sys.stderr.write("error: Index out of range \n")
+        sys.exit(-1)
+
+
+def get_archived_pvs(storage_name):
+    archived_pvs = {}
+    mntdir = os.path.join(HOSTVOL_MOUNTDIR, storage_name)
+    for pvc in yield_pvc_from_mntdir(mntdir):
+        print(pvc)
+        if pvc is not None and "pvc" in pvc["name"]:
```
vatsa287: Added "pvc" here for debug. Will change it to "archived_" later.

csi/remove_archived_pv.py
```diff
@@ -0,0 +1,64 @@
...
+    for pvname, values in archived_pvs.items():
+
+        print(pvname, values)
+
+        # Check for mount availablity before deleting info file & PVC
+        mntdir = os.path.join(HOSTVOL_MOUNTDIR, storage_name)
+        retry_errors(os.statvfs, [mntdir], [ENOTCONN])
+
+        # Remove PV in stat.db
+        update_free_size(storage_name, pvname, values["size"])
```
vatsa287: If PR #758 goes through then will keep as pvname else will replace("archived_", ""). Sent #758 to maintain uniformity of names across Info file, PVC & stat.db.

cli/kubectl_kadalu/remove_archived_pv.py
leelavg: - From Csi at line no. 18, 20 --> didn't get this.
- As you mentioned in slack, you aren't getting any values to `resp`? most probably there are multiple levels of exit codes here, need to re-check
- What's happening when you allow tty in kubectl command, i.e., `-t`?

cli/kubectl_kadalu/remove_archived_pv.py
```diff
@@ -0,0 +1,107 @@
...
+        # Validate method expects None when 'storage' not found.
+        return None
+
+
+def request_pv_delete(args):
+
+    cmd = utils.kubectl_cmd(args) + [
+         "exec", "kadalu-csi-provisioner-0", "-c", "kadalu-provisioner", "-nkadalu",
+         "--", "python3",
+         "-c", "'/kadalu/remove_archived_pv.py %s'" %(args.name)
```
leelavg: - is it trying to remove all the archive PVs in a pool?
- what extra `args` are being supplied to `utils.kubectl_cmd(args)`?

csi/remove_archived_pv.py
```diff
@@ -0,0 +1,64 @@
...
+        shutil.rmtree(os.path.dirname(info_file_path))
+
+        # Delete PVC
+        pvc_path = os.path.join(mntdir, values["path_prefix"], pvname)
+        shutil.rmtree(os.path.dirname(pvc_path))
+
+
+def main():
+    storage_name = get_names()
+    archived_pvs = check_archived_volumes(storage_name)
```
leelavg: - is `check_archived_volumes` already defined somewhere?

vatsa287: Yea have renamed this method to get_archived_pvs()

cli/kubectl_kadalu/remove_archived_pv.py
vatsa287: This is common across all such cmds across CLi. Most probably it has `kubectl`.

vatsa287: Yes in a given storage-pool. Will add select PVC functionality later.

csi/remove_archived_pv.py
```diff
@@ -0,0 +1,87 @@
...
+                        help="name of archived pvc belonging to specified storage-pool")
+
+    args = parser.parse_args()
+
+    archived_pvs = get_archived_pvs(args.name, args.pvc)
+    if archived_pvs == -1:
+        sys.exit()
+
+    if archived_pvs is not None:
+        sys.stdout.write("Found archived PVCs at storage pool %s" % args.name)
```
vatsa287: Used bash to call this file from CLI, made use of argparse to capture cmd line args. Now able to get stdouts and stderrs from CSI -> CLI / CSI container itself.

cli/kubectl_kadalu/remove_archived_pv.py
```diff
@@ -0,0 +1,110 @@
...
+    """ Send PVC delete request to CSI"""
+
+    cmd = utils.kubectl_cmd(args) + [
+         "exec", "-it", "kadalu-csi-provisioner-0", "-c", "kadalu-provisioner", "-nkadalu",
+         "--", "bash",
+         "-c", "cd /kadalu; python3 remove_archived_pv.py %s" %(args.name)
+    ]
+
+    if args.pvc:
+        cmd = utils.kubectl_cmd(args) + [
```
leelavg: - you could just extend above `cmd` list, isn't it?

csi/remove_archived_pv.py
vatsa287: done.

cli/kubectl_kadalu/remove_archived_pv.py
```diff
@@ -0,0 +1,110 @@
...
+
+def validate(args):
+    """
+    Validate the storage of archived pvcs requested to be delete
+    is present in kadalu configmap or not.
+    Exit if not present.
+    """
+
+    storage_info_data = get_configmap_data(args)
+
```
leelavg, line 33: - it's better to add another validation that PVC name if supplied should start with `archived`?

vatsa287: Done. Added it in CSI to avoid having checks in two places, if the standalone script is to be run from CSI.

cli/kubectl_kadalu/remove_archived_pv.py
vatsa287: Done. Edited out the last element. Append would cause err.

leelavg (approved):
/lgtm

amarts (approved):

PR #758, opened 2022-02-02
Update PV name in stat.db file after archival

vatsa287:

### DIFF
--- a/csi/volumeutils.py
+++ b/csi/volumeutils.py
@@ -219,6 +219,20 @@ def update_free_size(hostvol, pvname, sizechange):
                 acc.update_pv_record(pvname, -sizechange)
 
 
+def update_pv_name(hostvol, pvname, old_pvname):
+    """Update the pvname in respective host volume's stats.db file after archival"""
+
+    mntdir = os.path.join(HOSTVOL_MOUNTDIR, hostvol)
+
+    # Check for mount availability before updating the free size
+    retry_errors(os.statvfs, [mntdir], [ENOTCONN])
+
+    with statfile_lock:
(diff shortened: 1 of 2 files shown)


leelavg (approved):
- /lgtm
- I know that this PR is for #755, so you need to think of upgrade scenario as well
- In #755 you can't assume that every archived PVC will have name prefix in sqlite3
- So every change you make should consider the existing running code and modified code in the PR
- Generally, we do either of two things:
1. Whenever this code runs on existing setup, we should be able to modify existing references/logic and should mention in one of the releases that a specific release is a stop release and users should upgrade to that specific release before updating to current release.
Example:
v1 --> `mkdir -p /tmp/extra/data`
v2 --> `mkdir -p /tmp/data` (in addition to this we should also have `mv /tmp/extra/data /tmp/data` before `mkdir -p /tmp/data` is hit in codepath)
v3 -->  should notify user that dir at `/tmp/extra/data` is deprecated and everyone should upgrade to v2 before upgrading to v3 and only code for `mkdir -p /tmp/data` (basically remove extra logic from v2 if that's adding extra time)
2. We'll have both the logics (as explained in above example) and decide on the codepath based on version
- You have the flexibility to decide on what you choose depending on the functionality

@aravindavk pls weigh in your thoughts about deprecating features/functionalities.

vatsa287:
Thanks for the review!
#755 Is now working without this PR. Used existing update_free_size() by passing trimmed("archived-") pvc name to it. 
This PR is having CI issues. If it has more downsides can consider to close this. Wanted to just have uniformity in names of pvc after archival.

leelavg:
Can be closed for now.

Issue #766, opened 2022-02-20
UnboundLocalError: local variable 'response' referenced before assignment in exporter.py


fastlorenzo:
The following block should be placed within the previous try/except otherwise `response` doesn't exist if an issue occurs:

https://github.com/kadalu/kadalu/blob/24a634a62949b4e3768cb05e51f209186ac39147/operator/exporter.py#L282-L290

Stack trace:
```yaml
operator-79647d85d-rk82d kadalu-operator [2022-02-20 13:08:16,802] ERROR [exporter - 276:collect_all_metrics] - Unable to reach the pod, displaying only default values  pod_name=kadalu-csi-nodeplugin-dg98w error=HTTPConnectionPool(host='10.42.3.135', port=8050): Read timed out. (read timeout=1)
operator-79647d85d-rk82d kadalu-operator INFO:     10.42.1.21:33614 - "GET /metrics HTTP/1.1" 500 Internal Server Error
operator-79647d85d-rk82d kadalu-operator ERROR:    Exception in ASGI application
operator-79647d85d-rk82d kadalu-operator Traceback (most recent call last):
operator-79647d85d-rk82d kadalu-operator   File "/kadalu/lib/python3.9/site-packages/uvicorn/protocols/http/httptools_impl.py", line 375, in run_asgi
operator-79647d85d-rk82d kadalu-operator     result = await app(self.scope, self.receive, self.send)
operator-79647d85d-rk82d kadalu-operator   File "/kadalu/lib/python3.9/site-packages/uvicorn/middleware/proxy_headers.py", line 75, in __call__
operator-79647d85d-rk82d kadalu-operator     return await self.app(scope, receive, send)
operator-79647d85d-rk82d kadalu-operator   File "/kadalu/lib/python3.9/site-packages/fastapi/applications.py", line 208, in __call__
operator-79647d85d-rk82d kadalu-operator     await super().__call__(scope, receive, send)
operator-79647d85d-rk82d kadalu-operator   File "/kadalu/lib/python3.9/site-packages/starlette/applications.py", line 112, in __call__
operator-79647d85d-rk82d kadalu-operator     await self.middleware_stack(scope, receive, send)
operator-79647d85d-rk82d kadalu-operator   File "/kadalu/lib/python3.9/site-packages/starlette/middleware/errors.py", line 181, in __call__
operator-79647d85d-rk82d kadalu-operator     raise exc from None
[...]

PR #767, opened 2022-02-20
Fix issue when the request doesn't complete

fastlorenzo:


fastlorenzo:
Fixes #766

operator/exporter.py
```diff
@@ -272,23 +272,23 @@ def collect_all_metrics():
             response = requests.get(
                 'http://'+ pod_details["ip_address"] +':8050/_api/metrics',
                 timeout=1)
+            if response.status_code == 200:
```
leelavg, line 275: - Ideally these statements will not raise an error if `response` is valid
- So, why not assign `None` to `response` before `try` and `return metrics` if `except` block is hit

Just a thought

vatsa287 (approved):

operator/exporter.py
```diff
@@ -272,23 +272,23 @@ def collect_all_metrics():
...
-            if "nodeplugin" in pod_name:
-                set_nodeplugin_data(response, metrics, pod_name, pod_details)
-
-            if "provisioner" in pod_name:
-                set_provisioner_data(response, metrics, pod_name, pod_details)
-
-            if "server" in pod_name:
-                set_server_data(response, metrics, pod_name, pod_details)
-
+
```
amarts:
```suggestion
```

amarts:
DCO failure is because of my edits! The original commit has the signed-off, so we will consider this.

@leelavg any thing else you are looking at?

leelavg:
no, all good!

PR #771, opened 2022-02-27
Updated Read me to contains commands to allow kadalu to run on OCP. 

ChrisPhillips-cminion:
This was tested on OCP 4.8

Signed-off-by: Chris Phillips <chris.phillips@uk.ibm.com>

### DIFF
--- a/README.md
+++ b/README.md
@@ -36,6 +36,19 @@ curl -s https://raw.githubusercontent.com/kadalu/kadalu/devel/extras/scripts/cle
 ```
 
 
+### OpenShift
+If you are are running Kadalu on OpenShift (OCP)  the following commands must be run to give priviledged roles to the service accounts. 
+```
+oc adm policy add-scc-to-user privileged -z kadalu-operator  -n kadalu
+oc adm policy add-scc-to-user privileged -z kadalu-csi-nodeplugin -n kadalu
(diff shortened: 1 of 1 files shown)


leelavg:
@ChrisPhillips-cminion 

- thanks for the PR!
- could you pls cross check the manifest that you installed. It should be suffixed with openshift in the releases page
- SCCs are already part of the manifest https://github.com/kadalu/kadalu/blob/24a634/manifests/kadalu-operator-openshift.yaml#L34-L63

amarts:
Can we merge it and then look into the points updated by @leelavg as a followup?

leelavg:
- No, we better wait for the response
- What I was thinking which lead to this scenario is, user might have installed Kadalu via kadalu-cli and it didn't picked up correct manifest for the platform (let's say, cli picked up generic manifest in place of openshift)
- User tried to add SCCs and kadalu started working
- We need to make changes to CLI to pick-up correct manifest if it's not doing it currently

hunter86bg:
@ChrisPhillips-cminion ,
why would you need to add those on Openshift ?
I'm running Kadalu on OKD4.10 and I never needed to set those .

leelavg:
- Based on above comment and also we do have separate manifests for Openshift Deployments, closing this PR
- Pls re-raise if required

PR #778, opened 2022-03-25
change dir name of operator to kadalu_operator

leelavg:
- when importing functions from operator pkg locally there's namespace
  clash with inbuilt python operator pkg
- changed the name to kadalu_operator in the least number of places possible


aravindavk:
I prefer `kadalu_operator` instead of `manager`.

leelavg:
> I prefer `kadalu_operator` instead of `manager`.

Sounds good, will refresh at the earliest.

aravindavk (approved):

leelavg:
Oops! Seems I only changed pr msg but not commit msg 😕

PR #793, opened 2022-04-06
Support for Private container registry

aravindavk:
Build and upload the required container images to Private container
registry. Generate the manifest files to use those custom container
hub location.

Also support using local yaml files using Kadalu Kubectl extension.

### DIFF
--- a/doc/using-with-private-container-registry.adoc
+++ b/doc/using-with-private-container-registry.adoc
@@ -1,0 +1,81 @@
+= Install Kadalu Operator and CSI Pods using local Yaml and Container images.
+
+== Build and upload Container images
+
+Build container images by running the following command.
+
+----
+make build-containers
+----
+
+Push the following container images to your private container registry.
+
+----
+kadalu/kadalu-csi:devel
+kadalu/kadalu-operator:devel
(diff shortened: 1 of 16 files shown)


amarts (approved):

doc/using-with-private-container-registry.adoc
```diff
@@ -0,0 +1,81 @@
+= Install Kadalu Operator and CSI Pods using local Yaml and Container images.
+
+== Build and upload Container images
+
+Build container images by running the following command.
+
+----
+make build-containers
```
leelavg, line 8: - maybe we were in a hurry and merged this PR, I think below would be correct way @aravindavk as the current doc will leave room for manual steps, like re-tagging builds and current `make release` is incorrect if we want to use different IMAGES_HUB

1. change existing DOCKER_USER to REPO_USER all across
2. Use IMAGES_HUB in `build-containers` and supply to `build.sh`
3. in `release` as well we need to supply IMAGES_HUB

aravindavk: I will update the steps to pull from the docker hub once we make a new release. Build container step and install of kubectl-kadalu can be removed after that.

PR #809, opened 2022-05-03
Add openshift monitoring label, fixes kadalu#808

hunter86bg:
Signed-off-by: hunter86bg <hunter86_bg@yahoo.com>

When gen_manifest.py is triggered, the resulting manifests/kadalu-operator-openshift.yaml is created  has no label for Prometheus monitoring which leads to Prometheus ignore Kadalu Metrics

This commit should lead to an updated "kadalu-operator-openshift.yaml" that will have the label which will prevent users that update

**Checklist**
- [ ] Documentation added
- [ ] Tests updated

### DIFF
--- a/templates/operator.yaml.j2
+++ b/templates/operator.yaml.j2
@@ -16,6 +16,11 @@ apiVersion: v1
 metadata:
   name: kadalu-csi-nodeplugin
   namespace: {{ namespace }}
+  {% if k8s_dist == "openshift" %}
+  labels:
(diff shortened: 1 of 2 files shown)


leelavg:
In https://github.com/kadalu/kadalu/blob/devel/tests/minikube.sh#L285, change from `tail -n +6` to `tail -n +8`

hunter86bg:
Giving it up

PR #826, opened 2022-05-24
watch on config map mount based on deployment

### LINKED ISSUE #817: Fix kadalu on nomad environment
This PR breaks kadalu from working in a Nomad environment. There is no file or directory named ..data in a nomad environment. I would propose to setup a recursive watch of /var/lib/gluster instead, and put a sleep 1 into the loop, because else it has the potential of becoming very CPU intensive.

_Originally posted by @dhpowrhost in https://github.com/kadalu/kadalu/issues/715#issuecomment-1120766134_

leelavg:
fixes: #817


leelavg:
- No document update for Nomad is required, however the same can be referred from https://github.com/leelavg/kadalu-nomad/tree/v0.8.15

amarts (approved):

PR #828, opened 2022-05-27
multiple fixes in CSI RPC methods wrt voltypes

leelavg:
controllerserver:
- while expanding the volume infer the voltype from stored pvc metadata
  from backend
- there's a possibility only checking for accessbility may not
  differentiate b/n subvol and virtblock
- expansion of virtblock is similar to rawblock expect the fs expansion
  and show the same in code
- expand on PVC path of virtblock not on mountdir

quota-crawler:
- use find rather than ls to start quota update of subvol dirs

volumeutils:
- use os.truncate rather than raw truncate to workaround subproccess's
  'TypeError: expected str, bytes or os.PathLike object, not int'
- do not run mkfs.xfs while expanding virtblock volume
- delete a pvc and info metadata from leaf dir upto non empty dir
  excluding voltype dirs (subvol/virtblock/rawblock)
- removing the empty leaf nodes should speed up metrics collection if
  there are multiple PVC CRUD operations happening in the cluster
- include rawblock volumes in search operations

tests:
- set all the PVC original size to 200Mi and expand to 300Mi in the test
- set allowVolumeExpansion to true for custom created virtblock storage
  classes
- fix test for app pods success state
- add tests for expansion and verification of no leaf nodes when all
  PVCs are removed

### DIFF
--- a/examples/sample-test-app1.yaml
+++ b/examples/sample-test-app1.yaml
@@ -12,7 +12,7 @@ spec:
     - ReadWriteOnce
   resources:
     requests:
-      storage: 300Mi
+      storage: 200Mi
 # -*- mode: yaml -*-
 ---
 kind: PersistentVolumeClaim
@@ -27,7 +27,7 @@ spec:
     - ReadWriteMany
   resources:
     requests:
-      storage: 500Mi
+      storage: 200Mi
 # -*- mode: yaml -*-
 ---
 apiVersion: v1
@@ -77,6 +77,7 @@ apiVersion: storage.k8s.io/v1
 metadata:
   name: kadalu.storage-pool-1-virtblock
 provisioner: kadalu
+allowVolumeExpansion: true
 parameters:
   storage_name: "storage-pool-1"
   pv_type: Block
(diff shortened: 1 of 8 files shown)


examples/sample-test-app1.yaml
```diff
@@ -151,4 +151,4 @@ spec:
     - name: csivol4
       persistentVolumeClaim:
         claimName: pv1-4
-  restartPolicy: OnFailure
+  retartPolicy: OnFailure
```
amarts: spelling mistake?

tests/minikube.sh
```diff
@@ -438,6 +465,9 @@ case "${1:-}" in
     # deploy and validate app pods on storage pools created as part of 'kadalu_operator' case
     deploy_app_pods
 
+    # expand PVCs online and check the final size
+    expand_pvcs
```
amarts: nice! Good test.

leelavg:
- The more time I spend on this PR and more bugs I'm finding
- Most probably I'll fix all the issues that I faced and when CI passes I may break this PR into multiple PRs

leelavg:
> I may break this PR into multiple PRs
- taking this back, I may not be able to invest more time and break into separate PRs
- everything is tested as best as possible and updated the changes

amarts (approved):
LGTM

aravindavk (approved):

PR #832, opened 2022-06-01
fix find command in quota-crawler script

leelavg:

### DIFF
--- a/csi/quota-crawler.sh
+++ b/csi/quota-crawler.sh
@@ -6,7 +6,7 @@ echo "Starting the quota crawler script"
 count=0
 while true; do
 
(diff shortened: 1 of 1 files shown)


leelavg:
- i tested this earlier but somehow slipped from #828

sac (approved):
@leelavg looks good to me. Just curious why do you need -printf '.' ?
I mean why not just:
```dirs=$(find $MOUNT_DIR/*/subvol -mindepth 1 -maxdepth 1 -type d 2>/dev/null| wc -l);```

leelavg:
@sac thanks for the review,
- that'd also work, I just want the count not the paths/files, so it'll just print dot `.` for each find and `wc -c` count chars
- micro optimisation 😅

leelavg:
- there might be relatively more chances that users may deploy kadalu during weekends for evaluation, so merging this PR.

Issue #250, opened 2020-05-14
[RFE] Include Kubectl Kadalu in Operator image


aravindavk:
Installing a new package specific to Kadalu is not straight forward compared to Kubernetes setup. Many people are familiar with YAML based installation but not `pip` based installation.

Include Kubectl kadalu inside Operator, so that those utilities can be used using kubectl exec commands.

```
alias kadalucli='kubectl exec -it $(kubectl get pods -n kadalu -o name | grep operator | sed 's#pod/##') -n kadalu /usr/local/bin/kubectl-kadalu '
```

Now use Kadalu helper utility without installing any code in the host machine.

For example,

```
kadalucli storage-add ...
```

Document update required are:

* Kadalu install using `kubectl create -f` command.
* Storage add using YAML
* Alternate helper section explaining the above approach.
* Separate Doc for kubectl-kadalu install and use.

Changes required to Kubectl kadalu when installed inside the operator are:

* Don't expose `install` command(Since Kadalu Operator is already installed)
* Change `-h` or `--help` to something else, since it conflicts with `kubectl` command.(May be `kadalucli help storage-add`)

aravindavk:
good to have and easy to implement(only `Dockerfile` changes)

vatsa287:
> Installing a new package specific to Kadalu is not straight forward compared to Kubernetes setup. Many people are familiar with YAML based installation but not `pip` based installation.
> 
> Include Kubectl kadalu inside Operator, so that those utilities can be used using kubectl exec commands.
> 
> ```
> alias kadalucli='kubectl exec -it $(kubectl get pods -n kadalu -o name | grep operator | sed 's#pod/##') -n kadalu /usr/local/bin/kubectl-kadalu '
> ```
> 
> Now use Kadalu helper utility without installing any code in the host machine.
> 
> For example,
> 
> ```
> kadalucli storage-add ...
> ```
> 
> Document update required are:
> 
>     * Kadalu install using `kubectl create -f` command.
> 
>     * Storage add using YAML
> 
>     * Alternate helper section explaining the above approach.
> 
>     * Separate Doc for kubectl-kadalu install and use.
> 
> 
> Changes required to Kubectl kadalu when installed inside the operator are:
> 
>     * Don't expose `install` command(Since Kadalu Operator is already installed)
> 
>     * Change `-h` or `--help` to something else, since it conflicts with `kubectl` command.(May be `kadalucli help storage-add`)

[...]

PR #840, opened 2022-06-12
Allow building images directly from repository

### LINKED ISSUE #839: [RFE]: Ability to build images directly from the repository
**Is your feature request related to a problem? Please describe.**
Currently it is not possible to build images directly from the repository as the Dockerfiles
require the `requirements.txt`s to be in the repo's root directory (which `build.sh` currently does).

**Describe the solution you'd like**
Remove the redundant need to move the `requirements.txt`s in the first place allowing
build tools to directly build from Containerfiles.

**Describe alternatives you've considered**
/

**Additional context**
I would like to build images myself for testing and development. However, I run Kubernetes on ARM64. Building images from my local machine (MacOS) for ARM64 does not seem to work (error at installing grpcio).
For quick in-cluster builds I like using `kaniko`.

Perhaps there is a hidden reason for moving the `requirements.txt`s into the repository's root directory!? If there is not, then getting rid of that redundant step would make building a bit (c)leaner.

Timoses:
This removes a redundant copying of "requirements.txt"s
during the container image build process
and instead directly references the requirements files
in the `/requirements` directory.

This change enables usage of build tools which directly
build images from clong a git repository (like kaniko).

Removes redundant copying of `requirement.txt` files and removes dependency on `build.sh` for building Containerfiles from the repository.

Fixes #839 

Please let me know if I missed a hidden purpose for why this copying takes place.

**Checklist**
- [ ] -Documentation added- Not required AFAIK
- [x] Tests updated

### DIFF
--- a/.github/workflows/on-pr-merge.yml
+++ b/.github/workflows/on-pr-merge.yml
@@ -40,18 +40,14 @@ jobs:
         with:
           python-version: '3.10'
 
-      - name: Copy all requirements files to root of the repo for building images
-        run: |
-          cp -r requirements/* .
-
       - uses: actions/cache@v2
         with:
           path: ${{ env.pythonLocation }}
-          key: ${{ env.pythonLocation }}-${{ hashFiles('ci_merge-requirements.txt') }}
+          key: ${{ env.pythonLocation }}-${{ hashFiles('requirements/ci_merge-requirements.txt') }}
 
       - name: Get the CLI build (kubectl-kadalu)
         run: |
-          python -m pip install --upgrade --upgrade-strategy eager -r ci_merge-requirements.txt -e .
+          python -m pip install --upgrade --upgrade-strategy eager -r requirements/ci_merge-requirements.txt -e .
           KADALU_VERSION=devel make cli-build
 
       - name: Builder Image
--- a/.github/workflows/on-pr-merge.yml
+++ b/.github/workflows/on-pr-merge.yml
@@ -47,7 +47,7 @@ jobs:
 
       - name: Get the CLI build (kubectl-kadalu)
         run: |
-          python -m pip install --upgrade --upgrade-strategy eager -r requirements/ci_merge-requirements.txt -e .
+          python -m pip install --upgrade --upgrade-strategy eager -r requirements/ci_merge-requirements.txt -e ./requirements
           KADALU_VERSION=devel make cli-build
 
       - name: Builder Image
--- a/.github/workflows/on-pr-submit.yml
+++ b/.github/workflows/on-pr-submit.yml
@@ -50,7 +50,7 @@ jobs:
 
     - name: Install dependencies pylint
       run: |
-          python -m pip install --upgrade --upgrade-strategy eager -r requirements/ci_submit-requirements.txt -r requirements/ci_merge-requirements.txt -e .
+          python -m pip install --upgrade --upgrade-strategy eager -r requirements/ci_submit-requirements.txt -r requirements/ci_merge-requirements.txt -e ./requirements
 
     - name: Run pylint
       run: make pylint
--- a/.github/workflows/on-pr-submit.yml
+++ b/.github/workflows/on-pr-submit.yml
@@ -43,18 +43,14 @@ jobs:
       with:
         python-version: '3.10'
 
-    - name: Copy all requirements files to root of the repo
-      run: |
-        cp -r requirements/* .
-
     - uses: actions/cache@v2
       with:
         path: ${{ env.pythonLocation }}
-        key: ${{ env.pythonLocation }}-${{ hashFiles('ci_submit-requirements.txt') }}-${{ hashFiles('ci_merge-requirements.txt') }}
+        key: ${{ env.pythonLocation }}-${{ hashFiles('requirements/ci_submit-requirements.txt') }}-${{ hashFiles('requirements/ci_merge-requirements.txt') }}
 
     - name: Install dependencies pylint
       run: |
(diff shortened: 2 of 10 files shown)


leelavg:
@Timoses thanks for the PR
- if this PR passes CI then we don't have any issue
- iirc, the same format didn't work earlier for me when I made similar changes (or I might've missed something 🤔)
- let's see

Timoses:
Only task which still copies is https://github.com/kadalu/kadalu/blob/devel/Makefile#L138
But it seems to do this independent of anything else.
Can't get `pip-compile` to run locally, so I didn't touch it.

amarts (approved):

Timoses:
@leelavg I noticed there is still a dependency left:
https://github.com/kadalu/kadalu/blob/devel/kadalu_operator/Dockerfile#L49
which is noted as makefile dependency here:
https://github.com/kadalu/kadalu/blob/devel/Makefile#L24

I could not find a reference to using `kubectl kadalu` in the operator's code. Could kubectl-kadalu perhaps even be removed from the Operator image?

leelavg:
> Only task which still copies
- yes, we use pip-compile to generate dependencies based on `setup.{py,cfg}`
- if we aren't copying that into root of the repo then we need to `cd` into that dir everytime
- maybe better not to target that in this pr

> Could kubectl-kadalu perhaps even be removed
- better not, this is the plugin that you generally get when you do `kubectl krew install kadalu`
- having this in operator is just a convenience for users who don't want to install CLI tool


> I noticed there is still a dependency left:
- unfortunately yes, I can think of having a var CLI_BUILD to yes (default) or no in Makefile
- or anything is fine which can maintain current relations between CI (pr/merge/release) and Makefile targets

Timoses:
> > Only task which still copies
> 
>     * yes, we use pip-compile to generate dependencies based on `setup.{py,cfg}`
> 
>     * if we aren't copying that into root of the repo then we need to `cd` into that dir everytime
> 
>     * maybe better not to target that in this pr

Let's leave that for now then. It does not impact the purpose of this PR.

> > Could kubectl-kadalu perhaps even be removed
> 
>     * better not, this is the plugin that you generally get when you do `kubectl krew install kadalu`
> 
>     * having this in operator is just a convenience for users who don't want to install CLI tool
> 
> 
> > I noticed there is still a dependency left:
> 
>     * unfortunately yes, I can think of having a var CLI_BUILD to yes (default) or no in Makefile
> 
>     * or anything is fine which can maintain current relations between CI (pr/merge/release) and Makefile targets

I created https://github.com/kadalu/kadalu/issues/843 .

I'm not sure how to solve this in a hacky/fixy way now:
* A extra build ARG in the operator's Containerfile ?
    * However, there is no conditional `COPY` instruction
* Using another Container build target? e.g. `prod` would include the CLI, but `no_cli` target would not? (`prod` target would be based on `no_cli` and just COPY the CLI.
* ...?

leelavg:
- Maybe
``` make
BUILD_CLI ?= yes
cli-build:
ifeq ($BUILD_CLI,"yes")
.....build the binary....
else
....touch cli/build/kubectl-kadalu...
endif
```
- and you can custom build, `BUILD_CLI=false make build-containers`
- I didn't check existing targets, so above is just a rough idea which may work, wdyt?

Timoses:
> * Maybe
> 
> 
> ```makefile
> BUILD_CLI ?= yes
> cli-build:
> ifeq ($BUILD_CLI,"yes")
> .....build the binary....
> else
> ....touch cli/build/kubectl-kadalu...
> endif
> ```
> 
>     * and you can custom build, `BUILD_CLI=false make build-containers`
> 
>     * I didn't check existing targets, so above is just a rough idea which may work, wdyt?

That doesn't really solve the issue. What the goal is:
```
git clone https://github.com/kadalu/kadalu
cd kadalu
docker build -f kadalu_operator/Dockerfile .
```
No extra step required.

I like the idea of creating an empty file though.
Perhaps simply adding an empty file `cli/build/kubectl-kadalu` to the repo works? Or it could contain a comment like
```
# Hack to provide direct building of container images without build step (See https://github.com/kadalu/kadalu/issues/839)
```

Would that be ok?
(It's not beautiful, but looks like the cleanest hack)

leelavg:
@aravindavk what's your opinion on 
https://github.com/kadalu/kadalu/pull/840#issuecomment-1155386214?

aravindavk:
> @leelavg I noticed there is still a dependency left: https://github.com/kadalu/kadalu/blob/devel/kadalu_operator/Dockerfile#L49 which is noted as makefile dependency here: https://github.com/kadalu/kadalu/blob/devel/Makefile#L24
> 
> I could not find a reference to using `kubectl kadalu` in the operator's code. Could kubectl-kadalu perhaps even be removed from the Operator image?

`kubectl-kadalu` installed inside the Operator to allow users to run `kubectl-kadalu` using `docker exec operator-.. kubectl-kadalu` without installing `kubectl-kadalu` in the host machine. Refer: https://github.com/kadalu/kadalu/issues/250

aravindavk:
CLI code is not having any external dependency. CLI build steps can be extracted into a shell script and Makefile can also use that script. Operator Dockerfile can copy CLI directory and run `build-cli.sh` inside Operator container. 

For now, I am ok with empty file hack. I can send PR separately for making CLI build changes.

leelavg:
> inside the Operator to allow users to run
- yep, I'm aware of it, thanks for confirmation

> For now, I am ok with empty file hack
- @timoses I think you can move forward now

Timoses:
Rebased onto current `devel` branch.

leelavg:
@Timoses one last rebase and I'll merge this.

Timoses:
> @Timoses one last rebase and I'll merge this.

done

leelavg (approved):

PR #848, opened 2022-06-16
Harden RBAC permissions

### LINKED ISSUE #838: [RFE]: Hardening RBAC rules
**Is your feature request related to a problem? Please describe.**
Currently, Kadalu basically receives permissions onto the complete cluster. For security minded folks this is an issue and inhibits adoption of Kadalu.

**Describe the solution you'd like**
1. Revise operator RBAC permissions:
    * These should be narrowed down as much as possible (e.g. using namespaced Roles where applicable, granting only those API permissions needed for its operation)
2. Move RBAC definitions from [csi template](https://github.com/kadalu/kadalu/blob/devel/templates/csi.yaml.j2) into the operator manifest:
    * Defining them in the template obfuscates their existence. Having RBAC in one place gives a better overview.
    * Operator otherwise requires permissions to create these roles which means it can grant itself any other permissions
3. Specify separate RBAC rules for each CSI components (e.g. have a 'kadalu-external-provisioner' (cluster)role for [External Provisioner](https://github.com/kubernetes-csi/external-provisioner/blob/master/deploy/kubernetes/rbac.yaml)) which are then bound to the CSI controller service account. This should give a better overview of where RBAC requirements come from.

**Describe alternatives you've considered**
Perhaps it would also be an option to provide all RBAC related resources in an extra manifest `rbac.yaml`?

**Additional context**
I can take a look into it if you like. I'd be happy about pointers.
So far, I have identified:

**Operator's RBAC requirements** which are basically defined in the operator's calls to the API and `lib_execute(kubectl...)` calls in the `main.py` and the relating templates.

**Kadalu-Provisioner** requires [secret access](https://github.com/kadalu/kadalu/blob/devel/csi/controllerserver.py#L277) (thanks Leela for the pointer).

**CSI external components**: 
* [External provisioner](https://github.com/kubernetes-csi/external-provisioner/blob/master/deploy/kubernetes/rbac.yaml):
    * [secrets](https://github.com/kubernetes-csi/external-provisioner/blob/master/deploy/kubernetes/rbac.yaml#L24): **no** as far as I can see  (I understand it would mean the StorageClass definition would point to secrets to be passed to the CSI plugin)
    * [`PUBLISH_UNPUBLISH_VOLUME`](https://github.com/kubernetes-csi/external-provisioner/blob/master/deploy/kubernetes/rbac.yaml#L53) capability: **[yes](https://github.com/kadalu/kadalu/blob/devel/csi/protos/csi.proto#L885)**
    * [leadership election](https://github.com/kubernetes-csi/external-provisioner/blob/master/deploy/kubernetes/rbac.yaml#L77): **No**, [only one CSI controller pod via sts](https://github.com/kadalu/kadalu/blob/devel/templates/csi.yaml.j2#L178)
* [Attacher](https://github.com/kubernetes-csi/external-attacher/blob/master/deploy/kubernetes/rbac.yaml):
    * [secrets](https://github.com/kubernetes-csi/external-attacher/blob/master/deploy/kubernetes/rbac.yaml#L37): **no**, storageclasses in `templates/` do not seem to support this.
    * [leadership election](https://github.com/kubernetes-csi/external-attacher/blob/master/deploy/kubernetes/rbac.yaml#L61): **no**
* [Resizer](https://github.com/kubernetes-csi/external-resizer/blob/master/deploy/kubernetes/rbac.yaml):
    * [secrets](https://github.com/kubernetes-csi/external-resizer/blob/master/deploy/kubernetes/rbac.yaml#L25): **no**
    * [leader election](https://github.com/kubernetes-csi/external-resizer/blob/master/deploy/kubernetes/rbac.yaml#L62): **no**

Let me know what you think.

Timoses:
Signed-off-by: Timoses <timoses@tutanota.com>

For production deployments Kadalu permissions should be restricted to only required permissions.

Fixes #838 

* CSI RBAC now comes with opearator manifests to
  prevent operator requiring extra permissions
  to deploy RBAC itself
* CSI external component RBAC according to documentation
* Operator RBAC reduced to only required permissions (referred to operator's `main.py`)

Tested this with a fresh cluster in Replica3 mode.
It should be vetted more closely in other scenarios to ensure all required permissions are granted.

**Checklist**
- [ ] Documentation added - Not required AFAIK (permissions should be what kadalu needs for any scenario, no user interaction required)
- [ ] Tests updated - Not required IMO


Timoses:
One thing I noticed:
The operator has cluster-wide access to KadaluStorage CRD.
Not sure if this makes sense. Shouldn't only KadaluStorages be regarded in its own namespace?

Timoses:
The `pod-exec` role/binding was removed completely. I can not find a reference which points into this resources making any sense.


- [ ] TODO: If `pod-exec` can be removed, probably should also remove the following references:

```
> grep -irn 'pod-exec' . --exclude-dir=manifests --exclude-dir=helm
./doc/upgrade.adoc:42:  clusterrole.rbac.authorization.k8s.io/pod-exec unchanged
./extras/scripts/cleanup:22:kubectl delete -nkadalu ClusterRole pod-exec
```

leelavg:
@Timoses thanks a lot for raising pr.

- a hack was employed in the tests to verify helm chart to some extent implicitly 😅
- if there are some changes to operator and nodeplugin templates corresponding lines should be made to helm templates

leelavg:
> The pod-exec role/binding was removed completely.
- woah, this may not work
- iiuc this is used for running exec commands on the cluster, we basically use `kubectl` rather than k8s api in the operator most of the places

Timoses:
> > The pod-exec role/binding was removed completely.
> 
>     * woah, this may not work
> 
>     * iiuc this is used for running exec commands on the cluster, we basically use `kubectl` rather than k8s api in the operator most of the places

Should still work with this: https://github.com/kadalu/kadalu/pull/848/files#diff-aaaecf9f995faabb1ef2e9901431c91717589dd4bb7dbe8e7f1bcced87a6fc36L195
?

Edit: This link is a bit nicer.. https://github.com/kadalu/kadalu/blob/ecd8e5a87da51832681b275afa5ce37772bbe1ff/templates/operator.yaml.j2#L210

Timoses:
>     * a hack was employed in the tests to verify helm chart to some extent implicitly 😅
>     * if there are some changes to operator and nodeplugin templates corresponding lines should be made to helm templates

Uh, that's hacky xD. I did try to just copy and paste, though I'm not sure it'll withstand the judgement of a `diff` call : P

leelavg:
- pls rebase, thanks for the patience.

helm/kadalu/charts/operator/templates/rbac.yaml
```diff
@@ -0,0 +1,207 @@
...
+    verbs: ["get", "list", "watch", "patch"]
+  - apiGroups: ["storage.k8s.io"]
+    resources: ["csinodes"]
+    verbs: ["get", "list", "watch"]
+  - apiGroups: ["storage.k8s.io"]
+    resources: ["volumeattachments"]
+    verbs: ["get", "list", "watch", "patch"]
+  - apiGroups: ["storage.k8s.io"]
+    resources: ["volumeattachments/status"]
+    verbs: ["patch"]
```
leelavg, line 44: - kadalu doesn't use attacher sidecar and same will be informed to orchestrator via csidriver obj
- pls remove these and I believe these weren't there in current yaml files as well 🤔

helm/kadalu/charts/operator/templates/rbac.yaml
```diff
@@ -0,0 +1,207 @@
...
+  - apiGroups: ["storage.k8s.io"]
+    resources: ["csinodes"]
+    verbs: ["get", "list", "watch"]
+  - apiGroups: [""]
+    resources: ["nodes"]
+    verbs: ["get", "list", "watch"]
+  # PUBLISH_UNPUBLISH_VOLUME capability
+  - apiGroups: ["storage.k8s.io"]
+    resources: ["volumeattachments"]
+    verbs: ["get", "list", "watch"]
```
leelavg, line 80: - kadalu doesn't implement PUBLISH_UNPUBLISH_VOLUME RPC, this piece can be removed

helm/kadalu/charts/operator/templates/rbac.yaml
```diff
@@ -0,0 +1,207 @@
...
+metadata:
+  name: kadalu-csi-external-attacher
+subjects:
+  - kind: ServiceAccount
+    name: kadalu-csi-provisioner
+    namespace: {{ .Release.Namespace }}
+roleRef:
+  kind: ClusterRole
+  name: kadalu-csi-external-attacher
+  apiGroup: rbac.authorization.k8s.io
```
leelavg, line 129: - not required

leelavg (changes requested):

helm/kadalu/charts/operator/templates/rbac.yaml
Timoses: I thought it was used since: https://github.com/kadalu/kadalu/blob/devel/templates/csi.yaml.j2#L224
@leelavg  Please recheck, otherwise I can remove it of course.

helm/kadalu/charts/operator/templates/rbac.yaml
Timoses: The reason I assumed it would is because:
* RPC definitions are full of it (e.g.: https://github.com/kadalu/kadalu/blob/devel/csi/protos/csi.proto#L885)
* https://github.com/kadalu/kadalu/blob/devel/csi/nodeserver.py#L133

helm/kadalu/charts/operator/templates/rbac.yaml
Timoses: See https://github.com/kadalu/kadalu/pull/848#discussion_r913154345

helm/kadalu/charts/operator/templates/rbac.yaml
leelavg: - Good find and I'm sure we don't need this
- Corresponding sidecar can also be removed 😀

helm/kadalu/charts/operator/templates/rbac.yaml
leelavg: A bit of background if you don't mind
1. Controller (Un)publish is different from Node (Un)publish
2. The former is optional and latter is mandatory 
3. Former is mostly used in cloud environments, let's say AWS gp2-csi, where before a volume is created the storage need to be attched to the node
4. I believe there's no clear demarcation b/n controller and nodeplugin rbac I'll take a re-look after this PR merge

helm/kadalu/charts/operator/templates/rbac.yaml
Timoses: Ok, I'll remove them from here. Opened a bug report regarding running sidecar: https://github.com/kadalu/kadalu/issues/861

helm/kadalu/charts/operator/templates/rbac.yaml
Timoses: Ok, So since this role is only used by the CSI provisioner (controller part), I'll remove this.

leelavg:
@Timoses 
- could you pls revert last three commits from this PR?
- although I know for sure we don't use controller (un)publish rpc's CI is failing due to rbac
- I don't want you to go back and forth anymore, I'll take a relook after merge of this PR in it's original form

Pls excuse.

Timoses:
> @Timoses
> 
>     * could you pls revert last three commits from this PR?
> 
>     * although I know for sure we don't use controller (un)publish rpc's CI is failing due to rbac
> 
>     * I don't want you to go back and forth anymore, I'll take a relook after merge of this PR in it's original form
> 
> 
> Pls excuse.

Sure, no probs.

Removed the following commits:
```
* 26e73a5 Also remove CSI external attacher clusterrolebinding 6 hours ago Timoses (timoses@tutanota.com)
* 926c89a Remove CSI Provisioner (UN)PUBLISH role rights (controller part) 6 hours ago Timoses (timoses@tutanota.com)
* 6d865d8 Remove csi external attacher roles 6 hours ago Timoses (timoses@tutanota.com)
```

leelavg (approved):
- Thanks for the PR!

PR #868, opened 2022-07-22
CSI Refactor

aravindavk:
* Code refactored for easy maintenance.
* Changed Hosting Volume to Pool
* New Pool class provides all the required abstractions
  for Pool management.
* New Class PV provides all abstractions related to PV
* Changed Type=External to `Mode=external` and `Mode=native`
  is default.
* Kadalu format is renamed to `single_pv_per_pool=true|false`
* Changed Volume to Pool in Log messages, code.
* Removed code duplications.
* Volume context of existing PVs will be unchanged. All new PVs
  will have new Volume context(Versioned). Old Volume context
  is upgraded on the fly while reading.

With PV and Pool class CSI integration code is now very simple
to understand. For example, PV Create:

```python
pv_req = PV.from_csi_request(request)

# Filter and select the Pools that satisfy the parameters
# provided in Storage Class
pools = Pool.list(filters=pv_req.request.parameters)

# Shuffle the selected Storage Pools
random.shuffle(pools)

# Mount and select any one pool based on the size availability
pv_req.mount_and_select_pool(pools)

# PV Create
pvol = PV.create(pv_req)
```

And to mount the PV(NodePlugin),

```python
# Parse the Volume Context
pvol = PV.from_volume_context(request.volume_context)

# Mount PV. Pool is automatically mounted.
pvol.mount(request.target_path)
```

Signed-off-by: Aravinda Vishwanathapura <aravinda@kadalu.tech>

### DIFF
--- a/csi/volumeutils.py
+++ b/csi/volumeutils.py
@@ -13,11 +13,11 @@ import time
 from errno import ENOTCONN
 from pathlib import Path
 
+import csi_pb2
 import xxhash
 from jinja2 import Template
 from kadalulib import (PV_TYPE_RAWBLOCK, PV_TYPE_SUBVOL, PV_TYPE_VIRTBLOCK,
                        CommandException, SizeAccounting, execute,
-                       get_volname_hash, get_volume_path,
                        is_gluster_mount_proc_running, logf, makedirs,
                        reachable_host, retry_errors)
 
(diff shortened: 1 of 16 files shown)


aravindavk:
Still working on many functions. Not yet usable.

leelavg:
- Not sure how I missed this PR (maybe draft prs will not ring slack i guess)
- Nevertheless, this alone `Changed Hosting Volume to Pool` clears a lot of confusion, thanks for finally taking this up
- Let me know if you want to offload anything from this PR

aravindavk:
> * Not sure how I missed this PR (maybe draft prs will not ring slack i guess)
> 
>     * Nevertheless, this alone `Changed Hosting Volume to Pool` clears a lot of confusion, thanks for finally taking this up
> 
>     * Let me know if you want to offload anything from this PR

Thanks. Mostly done with changes, still some more changes remaining. You can expect the working PR by tomorrow or by this week.

amarts:
```
event.go:282] Event(v1.ObjectReference{Kind:"PersistentVolumeClaim", Namespace:"default", Name:"pv3-1", UID:"34f58456-dfaf-4485-b367-9820266247f7", APIVersion:"v1", ResourceVersion:"910", FieldPath:""}): type: 'Warning' reason: 'ProvisioningFailed' failed to provision volume with StorageClass "kadalu.storage-pool-3": rpc error: code = Unknown desc = Exception calling application: 'Pool' object has no attribute 'is_external'
```
Is it the reason for failure?

csi/volumeutils.py
```diff
@@ -1537,3 +1674,64 @@ def yield_list_of_pvcs(max_entries=0):
...
+
+    return None
+
+
+def upgraded_volume_context(ctx):
+    """
+    Upgrade the Volume Context with propper field names
+    """
+    # Already using the upgraded version
+    if ctx.get("version", "1") >= LATEST_VOLUME_CONTEXT_VERSION:
```
amarts, line 1745: File "/kadalu/volumeutils.py", line 1723, in upgraded_volume_context
    if ctx.get("version", "1") >= LATEST_VOLUME_CONTEXT_VERSION:
TypeError: '>=' not supported between instances of 'str' and 'int'

lib/kadalulib.py
```diff
@@ -385,25 +400,6 @@ def exit_gracefully(self, _signum, _frame):
         """When SIGTERM/SIGINT received"""
         self.terminating = True
 
-    # noqa # pylint: disable=no-self-use
```
leelavg, line 388: - I'd say better declare this as `@staticmethod` as this'll not be used out of this `Class`?

server/glusterfsd.py
```diff
@@ -125,19 +125,19 @@ def create_and_mount_brick(brick_device, brick_path, brickfs):
                     fstype=brickfs,
                     device=brick_device,
                     ))
-            except CommandException as err:
-                if "appears to contain an existing filesystem" not in err.err:
+            except CommandException as err1:
```
leelavg, line 128: - Maybe just `err=None` in line 120?

server/shd.py
```diff
@@ -48,7 +48,7 @@ def generate_shd_volfile(client_volfile, volname, voltype):
             if data["bricks"][(i * count)].get("decommissioned", "") != "":
                 decommissioned.append(brick_name)
 
-    data['decommissioned'] = "" if decommissioned == [] else ",".join(decommissioned)
+    data['decommissioned'] = ",".join(decommissioned) if decommissioned else ""
```
leelavg, line 51:
```suggestion
    data['decommissioned'] = ",".join(decommissioned)
```

csi/volumeutils.py
```diff
@@ -27,962 +27,1361 @@
...
 
 
-def filter_node_affinity(volume, filters):
+def options_based_pool_mount_suffix(opts, mnt_opts):
     """
-    Filter volume based on node affinity provided
+    If additional Options provided via Storage Class then
+    generate a suffix that can be used while mounting the Pool.
+    This allows to have multiple mounts of the same Pool with
+    different Volfile Options or Mount Options.
```
leelavg, line 62: Is `Volfile Options == at gluster server` and `Mount Options == at gluster client`?

csi/volumeutils.py
```diff
@@ -27,962 +27,1361 @@
...
+    if opts != "":
+        all_opts += opts
+    if mnt_opts != "":
+        all_opts += mnt_opts
 
-        # Volume is not from the requested node
-        if node_name != volume["bricks"][0]["kube_hostname"]:
-            return None
+    if all_opts != "":
+        return "_" + xxhash.xxh64_hexdigest(all_opts)
```
leelavg, line 71: - does the `opts` and `mnt_opts` will always be in order for hexdigest to work properly?

csi/volumeutils.py
```diff
@@ -27,962 +27,1361 @@
...
-        return None
+    pvhash = xxhash.xxh64_hexdigest(pvname)
+
+    return f"{pvtype}/{pvhash[0:2]}/{pvhash[2:4]}/{pvname}"
+
 
-    return volume
+def pv_abspath(mountpoint, pvpath, mount_suffix=""):
+    """Absolute Path of the PV including Pool mount path"""
+    return f"{mountpoint}{mount_suffix}/{pvpath}"
```
leelavg, line 88: - in here `mountpoint` itself will be changed due to the `suffix`?
- I didn't quite understood, what `mountpoint` will be, will it be `f'{POOL_MOUNTDIR}/{POOL_NAME}/'`?

csi/volumeutils.py
```diff
@@ -27,962 +27,1361 @@
...
+    makedirs(os.path.dirname(mountpoint))
+    Path(mountpoint).touch(mode=0o777)
+    execute(MOUNT_CMD, "--bind", loop, mountpoint)
+    return True
 
-    return volume
 
+# noqa # pylint: disable=too-many-public-methods
+# noqa # pylint: disable=too-many-instance-attributes
+class PV:
```
leelavg: - `PV` looks too terse, if possible can we use `PersistentVolume` itself?

csi/volumeutils.py
```diff
@@ -27,962 +27,1361 @@
...
+    """
+    PV object
+    """
+    def __init__(self, **kwargs):
+        self.name = kwargs.get("name", "")
+        self.type = kwargs.get("type", "")
```
leelavg, line 176: - `type` seems quite generic and may resemble Python `type`, can we change it to `pvtype`?
- I do know this is under `PV` class and can be attributed to `PV` only, so this is only a thought

csi/volumeutils.py
```diff
@@ -27,962 +27,1361 @@
...
+        return pvol
 
-    # Create a file with required size
-    makedirs(os.path.dirname(volpath_full))
-    logging.debug(logf(
-        f"Created {pvtype} directory",
-        path=os.path.dirname(volpath)
-    ))
+    @classmethod
+    def create(cls, req):
```
leelavg, line 340: - not able to pinpoint exactly the issue however there seems to be something wrong here
- when you create a `PV` you are not linking it to a `Pool` however you are just storing a string version of `Pool` and calling methods off of that, like `check_mount_availability(pvol.pool.mountpoint)`
- In above `mountpoint` is an attr of `pool` but I don't see the linking of `PV` obj to `Pool` obj
- Maybe we need to maintain a global dict of `Pool` objects and link them while creating `PV`(s)
- Pls correct me if i'm missing something (if my deduction is correct, there should be multiple pylint issues which I don't see)

leelavg (commented):
- review pending for `csi/*` and reviewed only half of `volumeutils.py`
- will do the remaining at a later time
- breaking this PR into multiple commits would've been good

csi/volumeutils.py
aravindavk: Volfile Options are also Client side options (Equivalent to gluster volume options but here same each Storage class can define different options for the same Pool), used inside the Volfile. Mount options are passed to the mount command while mounting it. 
- For external Gluster Volume, Kadalu don't have control on Volfile generation, but external Gluster mounts can be tuned by providing the mount options. To change the Volfile option, use gluster volume set command externally. (Note: Any option set with this method is applicable to all PVs created in that Storage pool) 
- Native Kadalu Volumes, doesn't use the mount command. So mount options here is Void.

csi/volumeutils.py
aravindavk: This can be fine-tuned later. For now, if two Storage classes provide options like `log-level=DEBUG,opt1=val1` and `opt1=val1,log-level=DEBUG` will create two mounts in respective CSI node plugin pod.

csi/volumeutils.py
aravindavk: CSI provisioner doesn't require multiple Storage pool mounts even if Options are different. Only node plugins creates multiple mounts when Application mounts the PV. 

In CSI node plugin, if the Options hash is H1 then the Pool mount becomes `/mnt/<pool>_H1` to differenciate the Pool mount used by other apps that doesn't modified the Volume options(`/mnt/<pool>`)

> I didn't quite understood, what mountpoint will be, will it be f'{POOL_MOUNTDIR}/{POOL_NAME}/'

Yes. `f'{POOL_MOUNTDIR}/{POOL_NAME}_{GLUSTER_VOLNAME}/'` if it is external Pool and pool name is not same as external gluster volume.

csi/volumeutils.py
aravindavk: Ack

csi/volumeutils.py
aravindavk: Ack. Will change the kwargs argument to `pvtype` but retain the PV class attribute.

csi/volumeutils.py
aravindavk: Look for `mount_and_select_pool` method to see PV and pool binding.

https://github.com/kadalu/kadalu/pull/868/files#diff-850490ad0b2666d1ea49b540576d0e6bb6525baf9e7b87d84fd841f80758ec4eR709-R718

`req.pool` is already populated by the above method. Check the example given in the commit message https://github.com/kadalu/kadalu/pull/868#issue-1315188973

lib/kadalulib.py
aravindavk: Ack

server/shd.py
aravindavk: Ack

csi/volumeutils.py
leelavg: ah! didn't reach there yet I guess, will probably cover some of those in next pass.

amarts:
Nice, finally passing all CI. 

My recommendation is, start merging this to another branch, and continue with Moana integration there with multiple PRs on that particular branch. Helps to review well, and be aware of what changes happened for integration. That way, it would be easier when everything is merged back to `devel`.

aravindavk:
Moved the code to Moana branch. Closing this PR.

PR #874, opened 2022-08-14
handle_modified can create volumes

blampe:

If a KadaluStorage instance was created with an invalid configuration (for example 3 bricks for Replica2) it would not create an entry in the kadalu-info CM.

If the user then edits the KadaluStorage to make it valid, handle_modify would not see it in kadalu-info and would refuse to do anything. However, given that the CRD is now valid this can be treated as the first time we've seen the resource.

Fixes #<issue number>

**Checklist**
- [ ] Documentation added
- [ ] Tests updated


leelavg:
- good catch! thanks for the PR.

leelavg (approved):

aravindavk (approved):

Issue #869, opened 2022-07-25
[Bug]: how to monitor pvc usage with kadalu in prometheus


ghost:
Hello dear community!

1. `Kadalu` installed via `Helm` (`operator` + `csi-nodeplugin`)
2. Connected to external `GlusterFS` (via `KadaluStorage`)
3. No significant metrics are available (especially about `PVC` usage)

Kubernetes version: `1.23.7`
Kadalu version: `0.8.15`

Please, could you help me how to monitor `PVC` usage?

Thanks in advance!

ghost:
@amarts @aravindavk @leelavg could you please help?

leelavg:
1. Kadalu currently exports these metrics https://github.com/kadalu/kadalu/blob/ed620e12493bd87f0f91beb980de7eb6c2a547dd/kadalu_operator/metrics.py#L3-L23
2. Sample from the CI
``` json
{
  "operator": {
    "memory_usage_in_bytes": 115666944,
    "cpu_usage_in_nanoseconds": 7084349100,
    "pod_name": "operator-578764959-wlvjq",
    "ip_address": "172.17.0.3",
    "pod_phase": "Running",
    "total_number_of_containers": 1,
    "number_of_ready_containers": 1,
    "containers": [
      {
        "container_name": "kadalu-operator",
        "is_ready": true,
        "is_started": true,
        "start_time": "2022-07-13T10:51:42Z"
      }
    ]
  },
  "storages": [
    {
      "name": "storage-pool-1",
      "type": "Replica1",
      "total_capacity_bytes": 12853444608,
      "free_capacity_bytes": 12532928512,
      "used_capacity_bytes": 320516096,
      "total_inodes": 6291456,
      "free_inodes": 6290431,
      "used_inodes": 1025,
      "pvc": [],
      "bricks": [
        {
          "brick_path": "/bricks/storage-pool-1/data/brick",
          "kube_hostname": "fv-az77-254",
          "node": "server-storage-pool-1-0-0.storage-pool-1",
          "node_id": "node-0",
          "host_brick_path": "",
          "brick_device": "/mnt/sda1/file1.1",
          "pvc_name": "",
          "brick_device_dir": "/mnt/sda1",
          "decommissioned": "",
          "brick_index": 0,
          "pod_phase": "Running",
          "memory_usage_in_bytes": 6192709632,
          "cpu_usage_in_nanoseconds": 1940439475400,
          "total_number_of_containers": 1,
[...]

Issue #852, opened 2022-06-17
[Bug]: High CPU usage in operator


TonyBogdanov:
**Describe the bug**
When the operator is running on a cluster with nodes that are frequently rotated, the operator goes into a state of high CPU usage the first time a large batch of nodes are added/removed, and remains there indefinitely.

In my particular case I have 20 ephemeral nodes that are provisioned in order to handle some jobs and then deleted about 10 minutes later. During that time the daemonset creates `csi-nodeplugin` pods on each node, which the operator talks to.

The `csi-nodeplugin` pods themselves behave normally all of the time, but the operator goes crazy and stays there.

Using `iftop` on the operator pod I can see a constant ~5mb/s transfer between the pod and `kubernetes.default.svc.cluster.local`. My personal best guess is that the operator connects to the newly created csi pods and does not clean up the connections when they die.

Maybe keeps trying to poll them or something like that?

**To Reproduce**
1. Install the kadalu operator.
2. Create 20 nodes and wait for the `csi-nodeplugin` daemonset to boot up pods.
3. Delete the 20 nodes.
4. The operator remains in a high CPU usage state until it is restarted.

**Expected behavior**
The operator should remain at a sane CPU usage level.

**Actual behavior**
The operator uses too much CPU after having a batch of nodes added and then removed.

**Environment:**

 - K8S_DIST: **kubernetes**
 - kubectl-kadalu plugin: **0.8.14**
 - pod/kadalu-csi-nodeplugin-2g9rz: **0.8.14**
 - pod/kadalu-csi-nodeplugin-68567: **0.8.14**
 - pod/kadalu-csi-nodeplugin-gtd2n: **0.8.14**
 - pod/kadalu-csi-nodeplugin-hcp56: **0.8.14**
 - pod/kadalu-csi-nodeplugin-hhdtj: **0.8.14**
 - pod/kadalu-csi-nodeplugin-jx4cb: **0.8.14**
 - pod/kadalu-csi-nodeplugin-qdv6q: **0.8.14**
 - pod/kadalu-csi-nodeplugin-zbtlz: **0.8.14**
 - pod/kadalu-csi-provisioner-0: **0.8.14**
 - pod/operator-6bcc99cff4-6q428: **0.8.14**
 - pod/server-storage-0-0: **0.8.14**
 - pod/server-storage-1-0: **0.8.14**
 - pod/server-storage-2-0: **0.8.14**

TonyBogdanov:
[...]

PR #879, opened 2022-08-22
another round of multiple fixes

leelavg:
csi:
- do not read pool & pvc details from nodeplugin for metrics collection
- ref: #852

manifests:
- add a new labels to operator to be able to get all pods which are part
  of kadalu using a single label

rbac:
- add missing rbac for listing persistentvolumes
- without this rbac the storageclasses weren't removed after pool
  deletion

dockerfile:
- install `procps` as part of building image
- for some reason `ps` isn't available in some of the component images,
  having `ps` will be quite handy for debugging rather than going into
  `/proc`

operator:
- only get the pods that are part of kadalu for metrics collection
- rename vars for better readability
- add empty `bricks` data for metrics collection of PVCs from external
  pools, ref: #869
- use same logic for finding presence of PVs from kadalu pools for both
  external and internal pools
- update configmap even in handle_added, since there's no finalizer pool
  CR can be removed for any reason and re-creation will not take into
  effect if we check for cm existence without updating
- without updating cm in handle_added, `volume_id` will never be used
  correctly incase of unintentional pool CR deletions

### DIFF
--- a/csi/exporter.py
+++ b/csi/exporter.py
@@ -41,10 +41,9 @@ def metrics():
     }
 
     if os.environ.get("CSI_ROLE", "-") == "nodeplugin":
-        pod_name_path = '/etc/hostname'
-        with open(pod_name_path, 'r') as pod_fd:
-            pod_name = pod_fd.read().strip()
-            data["pod"].update({"pod_name": pod_name})
+        # pool & pvc details can be read from provisioner
+        # avoid sending redundant data from nodeplugins
+        return data
 
     # Handle condition for no storage & PVC,
     # sometimes storage name is not shown at /mnt until server is mounted.
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Support custom kubelet dir
 - Container images can now be built directly from the repository without dependencies on previous build steps
 - Enable and correctly handle tolerations supplied in KadaluStorage CR
+- Multiple fixes around handling deletions and metrics (#879)
 
 ## [0.8.15] - 2022-06-16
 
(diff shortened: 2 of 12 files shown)


leelavg:
- All tested locally.

csi/exporter.py
```diff
@@ -41,10 +41,9 @@ def metrics():
     }
 
     if os.environ.get("CSI_ROLE", "-") == "nodeplugin":
-        pod_name_path = '/etc/hostname'
-        with open(pod_name_path, 'r') as pod_fd:
-            pod_name = pod_fd.read().strip()
-            data["pod"].update({"pod_name": pod_name})
+        # pool & pvc details can be read from provisioner
```
vatsa287, line 44: Is Pool & PVC details coming out of nodeplugin? I thought only pod details such as name, memory, cpu usage is coming out of it.
If this commented on, then even prometheus related data of nodeplugin must be commented out.

vatsa287:
Regarding high cpu usage, I think it is because of continous ping to pods which are either failed or not starting. 
Have to try it to stop fetching after sometime. But also keep in mind to fetch once it/(or any new pod) comes back up.

csi/exporter.py
leelavg: - yep but they are not being utilised by operator so it was just adding overhead

leelavg:
> Regarding high cpu usage, I think it is because of continous ping to pods which are either failed or not starting. Have to try it to stop fetching after sometime. But also keep in mind to fetch once it/(or any new pod) comes back up.
- the scraping happens intermittently, we aren't the once who is initiating requests, prom does that, shouldn't be a problem i guess

amarts (commented):
LGTM.

@aravindavk you want to have a look here? Otherwise, I am fine with merging this.

@leelavg thanks for this, and after the merge of other PRs, you may need a rebase now.

amarts (approved):

leelavg:
@aravindavk ptal, after this PR I think we can make a release.

amarts (approved):

aravindavk (approved):

PR #891, opened 2022-09-07
conditionally build and push base builder image

leelavg:
Gist:
- When specific conditions are met we are going to push base image while merging PR as well
- When specific conditions aren't met we are not going to build base image in every PR

Existing:
=========

on-pr-submit:
------------
1. unconditionally build base image for every pr and syncing of every pr
2. use the newly built image for component images

on-pr-merge:
------------
1. just build the base image for every image but no re-use in building component images

on-release-tag:
---------------
1. build and push base image which'll be used in building component images on-pr-merge

Current PR:
==========

on-pr-submit:
------------
1. build base image only if we are going for a release or builder Dockerfile is edited
2. in other cases pull the image

on-pr-merge:
------------
1. build and push base image only if we are going for a release or builder Dockerfile is edited
2. building and pushing of other component images will re-use this newly built image

on-release-tag:
---------------
1. build and push base image which'll be used in building component images on-pr-merge (no change)

### DIFF
--- a/.github/workflows/on-pr-merge.yml
+++ b/.github/workflows/on-pr-merge.yml
@@ -18,19 +18,15 @@ jobs:
   devel-tag-push:
     runs-on: ubuntu-latest
     steps:
-      -
-        name: Check out Code
+      - name: Check out Code
         uses: actions/checkout@v2
         with:
           fetch-depth: 2
-      -
-        name: Set up QEMU
+      - name: Set up QEMU
         uses: docker/setup-qemu-action@v1
-      -
-        name: Set up Docker Buildx
+      - name: Set up Docker Buildx
         uses: docker/setup-buildx-action@v1
-      -
-        name: Login to DockerHub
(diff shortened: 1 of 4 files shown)


.github/workflows/on-pr-merge.yml
```diff
@@ -50,14 +46,27 @@ jobs:
...
         timeout-minutes: 120
         with:
           context: .
           file: extras/Dockerfile.builder
           platforms: linux/amd64
-          push: false
+          push: true
           load: true
           tags: |
             docker.io/kadalu/builder:latest
```
amarts, line 72: If we are pushing this image, should this be 'devel' tag? or latest is fine?

latest we add for release only for now IMO.

leelavg: we need `latest` as in every component image we are referring the base image with `latest` tag

amarts (approved):

PR #893, opened 2022-09-07
optimize manifest generate using helm

iutx:
1. `common` charts is the stable version,  don't need to frequent change every version.
2. Instead of `sed` manifest, `sed` charts version first
- If version is `devel`, Charts version is still `0.0.0-0`, will generate manifest with `devel` image tag in `helm/kadalu/charts/common/templates/_version.tpl`.
image

- If version is release, will instead of default version and generate manifest with release version.
image

Fixes #<issue number>

**Checklist**
- [x] Documentation added
- [x] Tests updated

Signed-off-by: Asher Liu [root@viper.run](mailto:root@viper.run)

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -48,6 +48,9 @@ helm-manifest:
 	@echo ---------------------------------------------------------------------
 	@# Since we are using sub charts we can't percolate the version with ease
 	@# and so using 'sed' to replace the tag
+ifneq ($(KADALU_VERSION), devel)
+	@cd helm; grep -rln '0.0.0-0' | grep Chart.yaml | xargs -I file sed -i 's/0.0.0-0/${KADALU_VERSION}/g' file;
+endif
 	@helm show crds helm/kadalu > manifests/kadalu-operator${filename_suffix}.yaml
 	@echo "$$namespace" >> manifests/kadalu-operator${filename_suffix}.yaml
 	@helm template --namespace kadalu helm/kadalu \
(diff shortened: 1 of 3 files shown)


iutx:
I try to move sub charts to `helm/`, tree e.g. 
.
└── common
└── csi-nodeplugin (dependencies common)
└── kadalu  (dependencies csi-nodeplugin, operator)
└── operator (dependencies common)

But it didn't work out very well, need to execute `helm dependency update`, maybe all-in-one is the better structure(e.g. version<=0.8.4).

Each chart should have independent values.yaml and version management.

leelavg:
> maybe all-in-one is the better structure(e.g. version<=0.8.4).
- agreed, we took the current road due to a split in manifests
- however, we can re-look at unifying both and manipulate using values

iutx:
> > maybe all-in-one is the better structure(e.g. version<=0.8.4).
> 
> * agreed, we took the current road due to a split in manifests
> * however, we can re-look at unifying both and manipulate using values

👍 agreed, i have some simple ideas about the all-in-one structure.

tree ： standard helm chart
```shell
. kadalu
└ crds
└ templates
     |- _helpers.tpl
     |- NOTES.txt
     |- resources yamls
     └ ...
└ Chart.yaml
└ values.yaml
```

values.yaml
```yaml
image:
  registry: 'docker.io'
  repository: 'kadalu'
  pullPolicy: 'IfNotPresent'
kubeletDir: ''
kubernetesDistro: 'kubernetes'

operator:
  enabled: false
  verbose: 'no'

csi-nodeplugin:
  enabled: false
  verbose: 'no'
```
_helpers.tpl: common template helper function, enabled e.g.
> instead of subCharts condition, use template function
```tpl
{{/*
kaDalu operator enabled
*/}}
{{- define "kadalu.operator.enabled" -}}
{{/* may have some others condition */}}
{{ .Values.operator.enabled -}}
{{- end -}}
```

In resource template, wrapped in a enabled switch function, operator resource e.g.
```yaml
{{- if (include "kadalu.operator.enabled" . ) }}
...
{{- end }}
```

The manifest result of `helm template` not much different than it is now,  😣 only upgrading will be difficult for Helm users.

leelavg:
> The manifest result of helm template not much different than it is now, persevere only upgrading will be difficult for Helm users.
- Understood, it's better to simplify the process before we hit 1.0
~- I believe current PR might break release of helm chart because of version changes in subcharts and so I didn't merge it yet~
- As we have a working solution in devel, we aren't in a hurry, do you want to work on simplifying the charts (it's also fine if we can't make it to next release 😄)?
- Btw, thanks for helping in refining charts and the onus isn't on you if you don't want to simplify it further however your inputs will be helpful

iutx:
> * I believe current PR might break release of helm chart because of version changes in subcharts and so I didn't merge it yet

This PR is a simple optimization, it doesn't seem to affect the release. We can close this PR if not necessary, optimization is not must.

> * As we have a working solution in devel, we aren't in a hurry, do you want to work on simplifying the charts (it's also fine if we can't make it to next release 😄)?

I will try to simplify Chart in a new PR, i hope to be of some help in improving it. 😄 💪

iutx:
> * I believe current PR might break release of helm chart because of version changes in subcharts and so I didn't merge it yet

This PR is a simple optimization, it doesn't seem to affect the release. We can close this PR if not necessary, optimization is not must.

> * As we have a working solution in devel, we aren't in a hurry, do you want to work on simplifying the charts (it's also fine if we can't make it to next release 😄)?

I will try to simplify Chart in a new PR, i hope to be of some help in improving it. 😄 💪

PR #918, opened 2022-11-08
Bump up kubernetes python package

aravindavk:
Signed-off-by: Aravinda Vishwanathapura <aravinda@kadalu.tech>

### DIFF
--- a/kadalu_operator/main.py
+++ b/kadalu_operator/main.py
@@ -1000,11 +1000,6 @@ def main():
     """Main"""
     config.load_incluster_config()
 
-    # As per the issue https://github.com/kubernetes-client/python/issues/254
-    clnt = client.Configuration() #go and get a copy of the default config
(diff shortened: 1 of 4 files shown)


aravindavk:
Version not changed, need to run `gen requirements` I guess. I will update the PR

leelavg:
@aravindavk I'm not sure whether this works or not as there was a subtle bug in later version of k8s pkg that prevents accessing config from in-cluster the last time I checked.

Could you pls run `make gen-requirements` as the current CI using existing req and it passed.

amarts (approved):
The update of this package was pending from a long time. 

Good to get in. Also, with this and glusterfs codebase update, lets call this 0.9.x release, and then if things work out well, by Dec, before 2023, we are good for 1.0.0 IMO.

leelavg (approved):

PR #926, opened 2022-11-16
Apply tolerations from CR on server and nodeplugin pods

leelavg:
- thoroughly tested and might fix many asks from users

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,16 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ## [Unreleased]
 
 - Helm chart: fixed namespace management bug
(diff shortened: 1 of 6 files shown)


leelavg:
- will be updating the metrics docs in another PR

amarts (approved):
LGTM. Would be good if @aravindavk also takes a look.

aravindavk (approved):

fastlorenzo:
@leelavg this doesn't cover server pods as they are deployed using `StatefulSet` and not `DaemonSet`

leelavg:
@fastlorenzo I don't think so, tested this before raising PR.

Many I know what's missing from https://github.com/kadalu/kadalu/blob/devel/templates/server.yaml.j2#L26-L36?

Issue #959, opened 2023-02-08
[RFE]: Use kadalu_volgen library


aravindavk:
Kadalu Volgen is now available as a CLI tool/Library. Use that to generate the Volfiles required in this project.

https://github.com/kadalu/volgen

## Why?

- Volfile generation logic will be handled by the `kadalu_volgen` project.
- Supports GlusterFS compatible way of setting Options
- More features

## How?

- In server Pods, generate the below volfiles using the `kadalu_volgen`
    - Generate Client Volfiles (New change to use the Volfile server feature of Storage unit/Brick process)
    - Generate Storage Unit/Brick Volfile
    - Generate SHD volfile
    - Replace Shd volfile path with Volfile server
- In CSI pods,
    - Remove all the Volfile generation logic
    - Change the GlusterFS mount command: Use Volfile servers as storage unit server hostnames instead of Volfile path.
    - Use Storage class options as mount options only.
- In Operator,
    - Use `kadalu_volgen` to validate the option during Storage add.
    - Detect Option change by comparing existing options with the new options in CRD(`handle_modify`)
    - If Option changed, Exec to Server pods and re-generate all the Volfiles(Client, Shd, Brick/Storage Unit).
    - Send signal (`SIGHUP`) to running server brick process.

PR #966, opened 2023-03-06
csi: Use volfile server for fetching client-volfiles

vatsa287:

- This PR makes changes in CSI necessary to make use of volfile-server feature of brick processes,
where regeneration of client volfiles are not required, instead can fetch from any one of brick_processes.
Can add multiple brick_process to be notified as backup volfile servers. Which serves the volfiles stored in 
`/var/lib/kadalu/volfiles` inside server pods.
Can be accessed using `--volfile-server` option.

- Remove storage_options from mount flow. Later in upcoming PRs create a framework for storage_options to be configured at operator component.

Updates: #959 

**Checklist**
- [x] Documentation added : No user related information are expected from this PR.
- [x] Tests updated : Existing tests are enough for this PR.

Signed-off-by: Shree Vatsa N [vatsa@kadalu.tech](mailto:vatsa@kadalu.tech)

### DIFF
--- a/csi/nodeserver.py
+++ b/csi/nodeserver.py
@@ -9,7 +9,7 @@ import csi_pb2
 import csi_pb2_grpc
 import grpc
 from kadalulib import logf
-from volumeutils import mount_volume, unmount_volume
+from volumeutils import mount_glusterfs, mount_volume, unmount_volume
 
 HOSTVOL_MOUNTDIR = "/mnt"
 GLUSTERFS_CMD = "/opt/sbin/glusterfs"
@@ -84,6 +84,8 @@ class NodeServer(csi_pb2_grpc.NodeServicer):
             'g_options': options,
             'type': voltype,
         }
+
+        mount_glusterfs(volume, mntdir, True)
 
         if voltype == "External":
             logging.debug(logf(
--- a/csi/nodeserver.py
+++ b/csi/nodeserver.py
@@ -9,7 +9,7 @@ import csi_pb2
 import csi_pb2_grpc
 import grpc
 from kadalulib import logf
-from volumeutils import mount_glusterfs, mount_volume, unmount_volume
+from volumeutils import mount_volume, unmount_volume
 
 HOSTVOL_MOUNTDIR = "/mnt"
 GLUSTERFS_CMD = "/opt/sbin/glusterfs"
@@ -63,8 +63,6 @@ class NodeServer(csi_pb2_grpc.NodeServicer):
         gvolname = request.volume_context.get("gvolname", None)
         options = request.volume_context.get("options", None)
 
-        # Storage volfile options
-        storage_options = request.volume_context.get("storage_options", "")
         mntdir = os.path.join(HOSTVOL_MOUNTDIR, hostvol)
 
         pvpath_full = os.path.join(mntdir, pvpath)
@@ -76,8 +74,7 @@ class NodeServer(csi_pb2_grpc.NodeServicer):
             hostvol=hostvol,
             pvpath=pvpath,
             pvtype=pvtype,
-            pvpath_full=pvpath_full,
-            storage_options=storage_options
+            pvpath_full=pvpath_full
         ))
 
         volume = {
@@ -88,21 +85,13 @@ class NodeServer(csi_pb2_grpc.NodeServicer):
             'type': voltype,
         }
 
-        mountpoint = mount_glusterfs(volume, mntdir, storage_options, True)
-
         if voltype == "External":
             logging.debug(logf(
                 "Mounted Volume for PV",
                 volume=volume,
-                mntdir=mntdir,
-                storage_options=storage_options
+                mntdir=mntdir
             ))
             # return csi_pb2.NodePublishVolumeResponse()
-
-        # When 'storage_options' is configured mountpoint & volfile path change,
-        # Update pvpath_full accordingly.
-        if storage_options != "":
-            pvpath_full = os.path.join(mountpoint, pvpath)
 
         logging.debug(logf(
             "Mounted Hosting Volume",
--- a/csi/volumeutils.py
+++ b/csi/volumeutils.py
@@ -16,7 +16,8 @@ from kadalulib import (PV_TYPE_RAWBLOCK, PV_TYPE_SUBVOL, PV_TYPE_VIRTBLOCK,
                        CommandException, SizeAccounting, execute,
                        get_volname_hash, get_volume_path,
                        is_gluster_mount_proc_running, logf, makedirs,
-                       reachable_host, retry_errors, get_single_pv_per_pool)
+                       reachable_host, retry_errors, get_single_pv_per_pool,
+                       is_host_reachable)
 
 GLUSTERFS_CMD = "/opt/sbin/glusterfs"
 MOUNT_CMD = "/bin/mount"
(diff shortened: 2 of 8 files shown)


csi/volumeutils.py
```diff
@@ -940,82 +941,11 @@ def generate_client_volfile(volname):
...
-                    return
-
-    logging.debug(logf(
-        "Sent SIGHUP to glusterfs process",
-        volname=volname
-    ))
-    return
-
-
 def reload_glusterfs(volume):
```
aravindavk: Remove this function itself. Both native and external volfile servers handles notifying the connected clients when any option changes.

csi/volumeutils.py
```diff
@@ -1026,16 +956,14 @@ def reload_glusterfs(volume):
...
-        if not generate_client_volfile(volname):
+        if not has_configmap_changed(volname):
             return False
-        # TODO: ideally, keep the pid in structure for easier access
-        # pid = VOL_DATA[volname]["pid"]
-        # cmd = ["kill", "-HUP", str(pid)]
-        cmd = ["ps", "--no-header", "-ww", "-o", "pid,command", "-C", "glusterfs"]
+
+        pid = VOL_DATA[volname].get("pid", 0)
+        cmd = ["kill", "-HUP", str(pid)]
```
aravindavk: L959-976 can be removed

csi/volumeutils.py
```diff
@@ -1218,7 +1148,10 @@ def mount_glusterfs(volume, mountpoint, storage_options="", is_client=False):
                 raise err
         return mountpoint
 
-    generate_client_volfile(volname)
+    # Todo:
```
aravindavk: Remove this "TODO" also. Let's remove options completely from Storage class or PV create/mount flow. Let's keep only as Storage pool configuration.

aravindavk: Also remove the lines with Volfile modifications. That way we can have uniformity once NFS mount also introduced. Options can only be configured for Storage pools via operator.

leelavg:
hi @aravindavk,

- I tried reviewing this and previous volfile PR but couldn't quite understand the logic and forward looking statements which might be introduced in upcoming PRs (like NFS support)
- I'm particularly restricting myself not to get too much involved in kadalu ecosystem, as such could you pls point what (specific) other projects should be referred for all integrations coming into this project (kadalu:devel)?

thanks.

vatsa287:
Hi! @leelavg. 
In the previous PR of server component changes, we used [kadalu volgen library](https://github.com/kadalu/volgen]), instead of local volfile generation in server component. This was to keep all the volfile generation logic used in both kadalu-manager(moana) and Kadalu Kubernetes project as one. Also now since brick process are capable of serving volfiles to clients without glusterd, and volfile changes are handled by them, ex. SIGHUP. [Gluster FS PR](https://github.com/gluster/glusterfs/pull/3745).

This PR makes use of that feature of brick process, instead of creating volfiles again in CSI layer, we used `--volfile-server <host>` to fetch volfiles from server component, stored at `/var/lib/kadalu/volfiles/volname.vol`. And will be removing storage_options/volfile_options from here and moving the logic to be handled by operator, on storage-pool input.

Recently, we made changes to support NFS mount on kadalu_storage [issue](https://github.com/kadalu/moana/issues/65), can read [here](https://vatsa.hashnode.dev/kadalu-storage-is-now-integrated-with-nfs-ganesha) for better understanding. So with that changes done, `NFS` mount will be made as default mount rather than current `FUSE` based mount which would become optional. Changes made to NFS upstream to support kadalu storage is [here](https://review.gerrithub.io/c/ffilz/nfs-ganesha/+/547270), rationale being that volumes would be available in MAC ecosystem, would have to try if the nodeplugin issue on upgrade would be solved with this approach.

leelavg:
> nodeplugin issue on upgrade
- yep, it should be solved with kernel client, ie, NFS

Thanks a lot for the info, will go through the links.

csi/volumeutils.py
vatsa287: Removed `reload_glusterfs`.
In csi/main.py. `'reload_glusterfs` is used to reconfigure mounts. Should that be removed too?
L#54 csi/main.py

csi/volumeutils.py
vatsa287: Removed volume_options from mount flow.

csi/nodeserver.py
```diff
@@ -88,22 +85,14 @@ def NodePublishVolume(self, request, context):
             'type': voltype,
         }
 
-        mountpoint = mount_glusterfs(volume, mntdir, storage_options, True)
```
vatsa287, line 91: Removed mountpoint as it was an un-used.

aravindavk:
> hi @aravindavk,
> 
>     * I tried reviewing this and previous volfile PR but couldn't quite understand the logic and forward looking statements which might be introduced in upcoming PRs (like NFS support)
> 
>     * I'm particularly restricting myself not to get too much involved in kadalu ecosystem, as such could you pls point what (specific) other projects should be referred for all integrations coming into this project (kadalu:devel)?
> 
> 
> thanks.

Most of it already covered by @vatsa287

## Volfile management
Currently based on the Volume info, volfiles (That is config files for glusterfs process) are generated using the templates by the GlusterFS client itself. That is CSI Provisioner and CSI node plugins. Before calling the `glusterfs` command to mount the volume, it generates the Client volfile and then passes it to `glusterfs` as `-f <volfile>`. Using volfile path instead of volfile server option makes the mount static. If any changes to the Options or new bricks list the GlusterFS clients will not know automatically. So SIGHUP and other logics needed in client side to detect that volume option changed or new bricks added etc.

In Gluster standalone setup, Glusterd serves the Volfiles to the connected clients and also keeps the connected clients list so that Glusterd can notify all the connnected clients(mounts) to reload the Volfile when any option changes. Recently Amar added Volfile server functionality to the brick processes itself. So that Gluster mounts can work as is even Glusterd doesn't exists. Brick process itself can serve Volfile via RPC to the connected clients.  This is very great feature, following things are now possible without Glusterd

- Volfile generation only server side
- Regular `mount -t glusterfs` works with Kadalu
- Brick processes keeps the connected clients list so that Volfile reload notification is automatic.
- NFS and Samba integrations will work since both are dependent on Volfile server.

New Volgen library is implemented so that it can be used by both Moana and this project. All the Volfile generation logic is handled in one place and can be used by multiple projects.

With the above two changes(volfile server and volgen library), only server pods need to generate the volfiles. CSI code becomes lighter because Volfile generation is not required and both external Gluster and native Kadalu use the same mount command.

## Storage Options:
Earlier efforts was to support Storage options via storage class, and each Gluster mount in CSI node plugin can use these options and customize the volfiles accordingly. Implementing this will make the mounts static again without any auto notification. We can support only mount options only if the access is done via fuse.

New plan is to implement Options support only for Pools (May be in Operator). Storage class may get support for mount options. I will write about this in detail.

## Access protocols
Along with the Volfile server feature, Vatsa enhanced Gluster FSAL (NFS Ganesha plugin) to understand multiple Volfile servers and different ports than `24007`. With these features, we can now add support for NFS or Samba along with fuse.

Making NFS default is not yet sure, and we need not do it. Storage Class can decide which access protocols to use.

aravindavk:
@vatsa287 add default options here while generating the client volfile. https://github.com/kadalu/kadalu/blob/devel/server/serverutils.py#L88

```python
options = copy.copy(DEFAULT_OPTIONS)
options.update(client_data.options)

kadalu_volgen.generate(
        "/var/lib/kadalu/templates/client.vol.j2",
        data=client_data,
        options=options,
        output_file=client_volfile_path
)
```

```python
DEFAULT_OPTIONS = {
    "performance.client-io-threads": "off",
    "performance.stat-prefetch": "off",
    "performance.quick-read": "off",
    "performance.open-behind": "off",
    "performance.read-ahead": "off",
    "performance.io-cache": "off",
    "performance.readdir-ahead": "off"
}
```

aravindavk (approved):

Issue #913, opened 2022-10-31
App's FUSE mount goes to `Transport Endpoint not connected` 


mohammaddawoodshaik:
**Describe the bug**
By any chance if kadalu-provisioner gets restarted then all the application having GlusterFS mount will start displaying `TransportENdpoint is not connected` even after kadalu-provisioner is came backup. 

**To Reproduce**
1. Setup Kdalu infra in a single node and make sure deploy an app with Gluster mount in it.
2. Delete the csi-node-plugin pod
3. Provisioner pod will get re-scheduled automatically
4. Now inside the app you will see the `Transport endpoint not connected` issue, it never recovers from it.

**Expected behavior**
I want the mount point to get recovered automatically.

**Actual behavior**
Application mount is stuck in `Transport Endpoint` error.

**Environment:**
K8 version - 1.24
Kadalu - 0.8.17
Brick path - XFS file system
Single node setup - Replica1

leelavg:
@mohammaddawoodshaik 
- what was the state of nodeplugin when above issue is observed.
- workload/app is more related to the state of nodeplugin but not to the provisioner

mohammaddawoodshaik:
@leelavg 
Sorry for the confusion, it was node-plugin pod I have deleted to simulate the scenario - what will happen if FUSE process gets restarted?
But the problem I see here is after FUSE process comes-backup it is not repairing the errored POD mounts.

leelavg:
- I already mentioned this in slack thread https://kadalu.slack.com/archives/CLTPS540N/p1667395933939249?thread_ts=1667363649.871229&cid=CLTPS540N

```
 I'd actually ask for that recovery script if that can be shared in the first place :sweat_smile:
 No, most if not all CSI drivers running Fuse deamons has this issue. Ntg can be done (explained quite well here https://github.com/ceph/ceph-csi/issues/703#issuecomment-557520551)

Ack for the rest.
```

Compunctus:
Well, you should mark daemonset pods as OnDelete - so admin will be able to update one node at a time.

leelavg:
Ah! good one, it'll work for upgrades. However, this couldn't solve accidental nodeplugin reboots (which is rare though).

Neverthess, will update Nodeplugin manifests, thanks.

deba10106:
[...]

PR #986, opened 2023-04-14
Change nodeplugin upgrade strategy and introduce postStart hook

leelavg:
We had two known problems wrt nodeplugin, i.e, reboot and variation of reboot during upgrade. For upgrade our earlier response was to have admin deploy nodeplugin during a maintenance window and drain nodes which was using volumes. We didn't have any specific solution for reboot scenario other than asking deployments to scale down & up.

With the help of users feedback/suggestions, these are solved as below:
1. Set nodeplugin upgrade strategy as `OnDelete`, via this admin can selectively choose which nodeplugin to get upgraded and can delete post operator upgrade, the side effect is nullifying all the effort we put into separating the helm charts. However, having a correct solution is worth the effort rather than leaving it.
2. Use a postStart lifecycle hook on nodeplugin to restart gluster process, this needs a bit of enhancement, however as this is a tested sol going w/ it for now.

Credits:
1. @Compunctus from https://github.com/kadalu/kadalu/issues/913#issuecomment-1322495384
2. @sky-philipalmeida from https://github.com/kadalu/kadalu/pull/645#issuecomment-2082117053


leelavg:
- I saw this update in my local branch and probably didn't pushed earlier
- Will re-work on this, raising it here not to miss

Edit: Probably, I was aiming for a single chart and forgot the unfinished work

madoe:
:thumbsup:
Just saw this now. This change would be very beneficial for the upgrade process. I would be more than happy to see this merged.

leelavg:
Should also include https://github.com/kadalu/kadalu/pull/645#issuecomment-2082117053

aravindavk (approved):

PR #1030, opened 2023-11-17
Propagate error during failed mount when using external gluster volume.

### LINKED ISSUE #1029: [Need Help]:  Expect shared path for different pods across machine/nodes
Hey guys, I am trying to adopt kadalu with my existing External glusterFS server.

```yaml
---
kind: KadaluStorage
metadata:
  name: pv-ext
spec:
  type: External
  single_pv_per_pool: true             # Also tried false either. not working
  details:
    gluster_hosts:
      - my-gluster-ip        
    gluster_volname: app             
    gluster_options: log-level=DEBUG   

---
kind: PersistentVolumeClaim
apiVersion: v1
metadata:
  name: pv-ext
spec:
  storageClassName: kadalu.pv-ext
  accessModes:
    - ReadWriteMany
  resources:
    requests:
      storage: 200Mi
---
apiVersion: v1
kind: Pod
metadata:
  name: pod-ext
  labels:
    app: sample-app
spec:
  containers:
  - name: sample-app
    image: docker.io/kadalu/sample-pv-check-app:latest
    imagePullPolicy: IfNotPresent
    volumeMounts:
    - mountPath: "/mnt/pv"
      name: csivol
  volumes:
  - name: csivol
    persistentVolumeClaim:
      claimName: pv-ext
```

The yaml is basically same as `/example` and quite simple.
The PVC is being applied to two different namespaces. Let's say it `ns-a` and `ns-b`

But when I deploy two pods with pvc claimd in same namespace, the file under volume_path cannot be seen by each other, but either file is ok during pod delete & re-created. Below is my attempts:

1. The mount point is same. e.g. '/mnt/pv'
2. create podA under `ns-a`, kubectl exec -i podA touch /mnt/pv/a.txt
3. create podB under `ns-b`, cannot find file `a.txt` under '/mnt/pv'.
4. create file `b.txt` in podB.  Also, the file could not be seen inside podA. It seems isolated under two different paths inside actual glusterFs volume.

Can anybody help me with it?

liyuntao:
We now correctly propagate errors that can happen from the command during volume mount when using external gluster.  

Previously, these failed mount  
 error didn't bubble up and went to the upper caller, resulting in normal start-up/running pod and hide the problem.

Fixes #1029

### DIFF
--- a/csi/volumeutils.py
+++ b/csi/volumeutils.py
@@ -1100,8 +1100,8 @@ def mount_glusterfs_with_host(volname, mountpoint, hosts, options=None, is_clien
     try:
         execute(*command)
     except CommandException as excep:
-        if  excep.err.find("invalid option") != -1:
-            logging.info(logf(
+        if  excep.err.find("invalid option") != -1 or excep.err.find("unrecognized option") != -1:
+            logging.warning(logf(
                 "proceeding without supplied incorrect mount options",
                 options=g_ops,
                 ))
@@ -1109,13 +1109,13 @@ def mount_glusterfs_with_host(volname, mountpoint, hosts, options=None, is_clien
             try:
                 execute(*command)
             except CommandException as retry_err:
-                logging.info(logf(
+                logging.error(logf(
                     "mount command failed",
                     cmd=command,
                     error=retry_err,
                 ))
                 raise retry_err
-        logging.info(logf(
+        logging.error(logf(
             "mount command failed",
             cmd=command,
             error=excep,
--- a/csi/volumeutils.py
+++ b/csi/volumeutils.py
@@ -1108,19 +1108,19 @@ def mount_glusterfs_with_host(volname, mountpoint, hosts, options=None, is_clien
             command = cmd + [mountpoint]
             try:
                 execute(*command)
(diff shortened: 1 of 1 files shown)


amarts (commented):
LGTM

liyuntao:
Hi @amarts,

Just added a commit covering another line that should also raise an error near retrying logic using the fallback no-option mount command.

During the manual test, I noticed that there was another kind of invalid option message(sample below), so I added this coverage in `if` statement. All two commits are properly rebased.

```bash
$ /usr/sbin/glusterfs --process-name fuse --debug --volfile-id invalid-gfs-vol --volfile-server 192.168.1.5 --lag-level=INFO /home/myuser/my_test_folder

/usr/sbin/glusterfs: unrecognized option '--lag-level=INFO'
Try `glusterfs --help' or `glusterfs --usage' for more information.
```

amarts:
@liyuntao you need to use `-s` while creating the commit. (for passing DCO).

liyuntao:
> @liyuntao you need to use `-s` while creating the commit. (for passing DCO).

Thanks bro, your advice saved my day.

PR #1057, opened 2024-04-09
Implement list-before-watch strategy for KaladuStorage

### LINKED ISSUE #1056: [Bug]: Frequent exceptions in Operator: "Expired: too old resource version: 1058643076 (1578740144)"
About every hour the operator throws the following exception:

```
  File "/kadalu/main.py", line 1080, in <module>
   main()
  File "/kadalu/main.py", line 1069, in main
    crd_watch(core_v1_client, k8s_client)
  File "/kadalu/main.py", line 911, in crd_watch
    watch_stream(core_v1_client, k8s_client)
  File "/kadalu/main.py", line 884, in watch_stream
    for event in k8s_watch.stream(crds.list_cluster_custom_object,
  File "/kadalu/lib/python3.10/site-packages/kubernetes/watch/watch.py", line 182, in stream
    raise client.rest.ApiException(
kubernetes.client.exceptions.ApiException: (410)
Reason: Expired: too old resource version: 1058643076 (1578740144)
```

This results in the regeneration of the kadalu-info configmap and in turn the restart of kadalu-provisioner and kadalu-nodeplugin (will be prevented at the moment due to bug #1053).

At the moment we have 21 external gluster volumes connected (`single_pv_per_pool=true`) some of them for nearly a year. Apparently the KadaluStorages, as reported back from the Kubernetes-API, are not sorted by resourceVersion but by name. So the resourceVersion of one of our oldest volume will be last reported when starting to watch for events:
```
> curl "http://127.0.0.1:8001/apis/kadalu-operator.storage/v1alpha1/kadalustorages?watch=1"
{"type":"ADDED","object":{...,"name":"aqg","namespace":"aqg","resourceVersion":"1058610339",...}
{"type":"ADDED","object":{...,"name":"bqh","namespace":"bqh","resourceVersion":"1058618634",...}
...
{"type":"ADDED","object":{...,"name":"hqg","namespace":"hqg","resourceVersion":"1331005221",...}
...
{"type":"ADDED","object":{...,"name":"vqg","namespace":"vqg","resourceVersion":"1058627852,...}
{"type":"ADDED","object":{...,"name":"vqh","namespace":"vqh","resourceVersion":"1058643076",...}
```

Currently the last reported resourceVersion will be used for a reconnect if the API-Connection is disrupted in any way, resulting in the forementioned exception:
```
>  curl "http://127.0.0.1:8001/apis/kadalu-operator.storage/v1alpha1/kadalustorages?watch=1&resourceVersion=1058643076"
{"type":"ERROR","object":{"kind":"Status","apiVersion":"v1","metadata":{},"status":"Failure","message":"too old resource version: 1058643076 (1578794510)","reason":"Expired","code":410}}
```

As mentioned in the [Kubernetes API documentation](https://kubernetes.io/docs/reference/using-api/api-concepts/#efficient-detection-of-changes) and the [comment in kubernetes-python-client](https://github.com/kubernetes-client/python/blob/e0234d3e7193aea4b4e8f487cab0f1228f4386d2/kubernetes/base/watch/watch.py#L132)
a better approach would be a list-then-watch approach, using the resourceVersion of the list/collection as a startpoint for the watch. This also has been mentioned in a discussion in the [kubernetes-python-client project](https://github.com/kubernetes-client/python-base/pull/131#issuecomment-502737581).

In upcoming versions of kubernetes a ```sendInitialEvents``` flag will be introduced (currently behind a feature gate), that could simplify the list-then-watch approach. 

*Environment:*

* Kadalu Version: [1.2.0]
* K8S_DIST: [kubernetes]
* external gluster with single_pv_per_pool=true

madoe:
The operator is using "watch" to get notified of changes to kadaluStorages. The resourceVersion used to restart the watch - if disrupted - could be unknown to Kubernetes, because 
* they will be discarded after time
* the initial events when starting watch are not ordered by resourceVersion

As mentioned in https://github.com/kubernetes-client/python/blob/e0234d3e7193aea4b4e8f487cab0f1228f4386d2/kubernetes/base/watch/watch.py#L132 it is better to use a list as initial state and use the resourceVersion of the list as starting point for "watch".

Fixes #1056

### DIFF
--- a/kadalu_operator/main.py
+++ b/kadalu_operator/main.py
@@ -880,7 +880,16 @@ def watch_stream(core_v1_client, k8s_client):
     """
     crds = client.CustomObjectsApi(k8s_client)
     k8s_watch = watch.Watch()
-    resource_version = ""
+    initial_list = crds.list_cluster_custom_object("kadalu-operator.storage",
+                                                   "v1alpha1",
+                                                   "kadalustorages")
+
+    for item in initial_list.get("items"):
+        handle_added(core_v1_client, item)
+
+    metadata = initial_list.get("metadata")
+    resource_version = metadata['resourceVersion']
+
     for event in k8s_watch.stream(crds.list_cluster_custom_object,
                                   "kadalu-operator.storage",
                                   "v1alpha1",


leelavg:
Hey, thanks for showing interest in the project and sending patches, I'll spend sometime by end of next week (if others doesn't pitch in by then) and go over your issues & PRs.

Forcing myself to take a look by stating above here 😅.

leelavg (approved):
fix makes sense, I haven't come across this bug as this can only be hit it there are no events for prolonged times and events may have fallen off the list.

I believe 1hr is the default limit for evicting events.

leelavg:
I've raised #1059 to fix the CI and the content in this PR shouldn't result in regression, merging it.

PR #1065, opened 2024-05-03
update on-pr-submit and dependents

leelavg:
Started work to only change Minikube Version and be done with it, however had to do these many changes as for Minikube to work we need to install extra tools due to deprecation of docker shim.

Only touched PR Submit workflows and may modify renaming in later PRs.

Changes:
1. Remove deprecated usage of save-ouput GH flow and use GITHUB_OUPUT
2. Use native cache from setup-python action rather than custom step
3. Remove minikube as we were testing against 3 year old kubernetes version (1.20) and replace minikube with k3d, bump kube version
4. Consolidate tests into a single file, a couple of years back main test file was refactored and as yearly maintainence except main test the environment is updated as we don't want to debug both tests & env failures at once
5. Since local docker is being used as pass through registry adding new steps to save and restore docker image cache


leelavg:
can look at https://github.com/leelavg/kadalu/blob/update-ci/tests/tests.sh for a clean file w/o changes.

PR #1086, opened 2024-11-21
Add argument for namespace

### LINKED ISSUE #1082: [Bug]: Helm chart allows installation in custom namespace, but Kadalu will only work in 'kadalu' namespace
## Description

* By using the Helm chart, Kadalu can be installed in an arbitrary namespace (e.g. setting `namespace: kadalu-system`)
* BUT Kadalu expects the namespace to be `kadalu`

## Expected Behavior

* Either Kadalu works with the namespace set in the Helm chart (preferred),
* or the installation fails with an error

## Actual Behavior

* Parts of Kadalu will not work, e.g. running kubectl-kadalu
    * `kubectl exec -i -t operator-XXX-YYYY -- kubectl-kadalu version`
    * Output: Error from server (Forbidden): pods is forbidden: User "system:serviceaccount:kadalu-system:kadalu-operator" cannot list resource "pods" in API group "" in the namespace "kadalu"

jochenseeber:
Allows accessing Kadalu installations installed in other namespaces than the default "kadalu".

The official Helm chart allows installing Kadalu in a custom namespace. However, the cli will not work because it assumes the namespace always is `kadalu`.

This commit adds a parameter `-n` to the cli to provide a custom namespace.

Fixes #1082.


amarts (approved):

amarts:
Looks like you would need to submit with `-s` in `git commit`

PR #1088, opened 2025-02-11
Fix PVC mounting issues when 1st server pod is unavailable

greenday9:
When the first server pod is unavailable, kadalu is unable to mount with errors like "DeadlineExceeded".  This is due to the is_server_pod_reachable code which will iterate over the same server pod multiple times.  The original code would try 1st server pod, timeout in 20, sleep 30 seconds, and then retry 4 times.  This results in up to 200 seconds to determine that 1st server pod  is not reachable before trying the next.  This is too long to wait before responding when the issue may be isolated to the 1st server pod.

Given that it appears that the purpose of is_server_pod_reachable is to see if _any_ server pod is available, I rewrote it to function semi-asynchronously.  Basically, it will attempt to connect to all server pods at once and see if any succeeds.  If none succeeded, then sleep 30 seconds and retry.

It reduces the blocking in is_server_pod_reachable to minimum to determine whether any server pod is reachable.  This allows PVC mounting to continue when 1st server is not reachable but other server pods are reachable.

Fixes mounting issues when 1st server pod is unavailable.

### DIFF
--- a/lib/kadalulib.py
+++ b/lib/kadalulib.py
@@ -120,9 +120,9 @@ def is_server_pod_reachable(hosts, port=24007, timeout=20):
                 sockets.append(sock)
 
         # Only care about output sockets (not input or exception)
-        readable, writable, exceptional = select.select([], sockets, [], timeout)
+        _, writable, _ = select.select([], sockets, [], timeout)
 
-        if not (readable or writable or exceptional):
+        if not writable:
             logging.info(logf(
                     "Failed to connect to any server pod..."
                 ))
--- a/lib/kadalulib.py
+++ b/lib/kadalulib.py
@@ -8,6 +8,8 @@ import sqlite3
 import subprocess
 import sys
 import time
+import select
+import errno
 
 import xxhash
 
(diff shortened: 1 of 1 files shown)


greenday9:
@leelavg , @amarts , @aravindavk , any thoughts on this change?

leelavg:
Possible to fix lint errors? I'll look at test failures when possible. Change lgtm, however, pls note I could only review & merge but can't cut out a release.

lib/kadalulib.py
```diff
@@ -87,23 +89,64 @@ def is_server_pod_reachable(hosts, port=24007, timeout=20):
...
+                        retry_count=retry_count+1
+                    ))
+
+            if connect_started:
+                sockets.append(sock)
+
+        # Only care about output sockets (not input or exception)
+        readable, writable, exceptional = select.select([], sockets, [], timeout)
+
+        if not (readable or writable or exceptional):
```
leelavg:
```suggestion
        # Only care about output sockets (not input or exception)
        _, writable, _ = select.select([], sockets, [], timeout)

        if not writable:
```
as ntg is supplied to select for rlist & xlist, the returned values will be of no significance afaik and could be discared, wdyt?

leelavg:
> at test failures when possible.
- seems the test suite is broken and I see no active development on the repo, so the suite may not be fixed.

lib/kadalulib.py
greenday9: Yes, I agree that would be better.

greenday9:
> Possible to fix lint errors? I'll look at test failures when possible. Change lgtm, however, pls note I could only review & merge but can't cut out a release.

pylint looks clean now.  I don't see any issues except the suppressed C0209(f-string).  I would still be in favor of merging even if we can't release although that is preferred.

leelavg:
@greenday9 could you pls provide a sample test output or an ack that you've tested the fix?

I'll merge after an ack as this doesn't create any security issues. I see no responses from others and so thinking to merge PRs which doesn't create security vulnerabilities (by looking at code).

PR #1090, opened 2025-03-14
Modify grep-pattern to match only on white space word borders

madoe:
**What this PR does / why we need it**:
`grep -w` matches on all word borders, not only white space. Grepping for a volume `example-pv` will therefore also report a match on `example-pv-suffix`, leading to "No such file or directory:"-errors when trying to unmount `example-pv` in kadalu-provisioner.

**Which issue(s) this PR fixes**:
Fixes #1077

**Special notes for your reviewer**:

**Checklist**
- [ ] Documentation added
- [ ] Tests updated
- [ ] Add an entry in the `CHANGELOG.md` about the changes.

PR #1092, opened 2025-03-20
Fix grep not matching volumes after PR #1090                                                                                           

madoe:
After implementing PR #1090 I noticed that grep will now never match to a volume, as the mountpoint in the output is always on the end of a line and will never have a trailing space. Changed the grep expression to match this expected output.

Fixes #<issue number>

**Checklist**
- [ ] Documentation added
- [ ] Tests updated

### DIFF
--- a/lib/kadalulib.py
+++ b/lib/kadalulib.py
@@ -67,7 +67,7 @@ def is_gluster_mount_proc_running(volname, mountpoint):
     cmd = (
         r'ps ax | grep -w "/opt/sbin/glusterfs" '
         r'| grep -w "\-\-volfile\-id %s " '
-        r'| grep -w -q "%s "' % (volname, mountpoint)
+        r'| grep -G -q "%s$"' % (volname, mountpoint)
     )
 
     with subprocess.Popen(cmd,
--- a/lib/kadalulib.py
+++ b/lib/kadalulib.py
@@ -66,8 +66,8 @@ def is_gluster_mount_proc_running(volname, mountpoint):
     """
     cmd = (
         r'ps ax | grep -w "/opt/sbin/glusterfs" '
-        r'| grep -w "\-\-volfile\-id %s " '
-        r'| grep -G -q "%s$"' % (volname, mountpoint)
+        r'| grep -- "--volfile-id %s " '
+        r'| grep -q "%s$"' % (volname, mountpoint)
     )
 
     with subprocess.Popen(cmd,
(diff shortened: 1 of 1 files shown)


leelavg:
Possible to provide the line that we run the grep on? I think the cmd can be simplified.

madoe:
An example line would be
`     51 ?        SLsl  26:02 /opt/sbin/glusterfs --process-name fuse -l /var/log/gluster/gluster.log --volfile-id test-volume-1 --volfile-server gluster-1.example.com --volfile-server gluster-2.example.com --volfile-server gluster-3.example.com --log-level=INFO /mnt/test-volume-1`

leelavg:
thanks, how about below
``` python
    cmd = (
        r'ps ax | grep -w "/opt/sbin/glusterfs" '
        r'| grep -- "--volfile-id %s" '
        r'| grep -q "%s$"' % (volname, mountpoint)
    )
```

``` sh
# cat /tmp/lines
51 ?        SLsl  26:02 /opt/sbin/glusterfs --process-name fuse -l /var/log/gluster/gluster.log --volfile-id test-volume-1 --volfile-server gluster-1.example.com --volfile-server gluster-2.example.com --volfile-server gluster-3.example.com --log-level=INFO /mnt/test-volume-1
51 ?        SLsl  26:02 /opt/sbin/glusterfs --process-name fuse -l /var/log/gluster/gluster.log --volfile-id test-volume-1-2 --volfile-server gluster-1.example.com --volfile-server gluster-2.example.com --volfile-server gluster-3.example.com --log-level=INFO /mnt/test-volume-1-2
51 ?        SLsl  26:02 /opt/sbin/glusterfs --process-name fuse -l /var/log/gluster/gluster.log --volfile-id t-test-volume-1 --volfile-server gluster-1.example.com --volfile-server gluster-2.example.com --volfile-server gluster-3.example.com --log-level=INFO /mnt/t-test-volume-1

# grep -- '--volfile-id test-volume-1' < /tmp/lines | grep -q /mnt/test-volume-1$ && echo matched || echo not-matched
matched

# ------ only matched test-volume-1
# grep -- '--volfile-id test-volume-1' < /tmp/lines | grep /mnt/test-volume-1$
51 ?        SLsl  26:02 /opt/sbin/glusterfs --process-name fuse -l /var/log/gluster/gluster.log --volfile-id test-volume-1 --volfile-server gluster-1.example.com --volfile-server gluster-2.example.com --volfile-server gluster-3.example.com --log-level=INFO /mnt/test-volume-1
```

Nevertheless, this is from your observation that mountpoint ends the line.

madoe:
Yes, that looks OK. However I'd prefer to add the trailing space at the second `grep` to eliminate any false matches caused by suffixes (the original error), even if the last `grep` should eliminate them.

PR #1094, opened 2025-03-31
* Fix Unmount issue, when gvolname and name differs

madoe:
**What this PR does / why we need it**:
The PR fixes two issues with external volumes:
1. gluster volumes were not unmounted, when gluster volume name and volume name differs
2. search_volume was looping through all available volumes, instead of just the one passed as parameter, mounting all available volumes in the process.

**Which issue(s) this PR fixes**:
Fixes #1083 

**Special notes for your reviewer**:

**Checklist**
- [ ] Documentation added
- [ ] Tests updated
- [ ] Add an entry in the `CHANGELOG.md` about the changes.

Issue #1083, opened 2024-10-02
[Bug]: kadalu-csi-provisioner-0 pod keeps external gluster volumes mounted using 11GB of memory


ramalhais-sky:
**Describe the bug**
kadalu-csi-provisioner-0 pod keeps external gluster volumes mounted using 11GB of memory

**To Reproduce**
1. Create KadaluStorage for several external gluster volumes
2. kubectl -n kadalu exec -it kadalu-csi-provisioner-0 -c kadalu-provisioner -- bash
3. Check number of mounts:
  ```
  root@kadalu-csi-provisioner-0:/# mount -t fuse.glusterfs|wc -l
  849
  ```
4. Check memory usage (sum VSS)
```
root@kadalu-csi-provisioner-0:/# ps auxww|grep glusterfs|awk '{print $6}'|while read; do MEMUSAGE=$(($MEMUSAGE+$REPLY)); echo MEMUSAGE=$MEMUSAGE; done | tail -1
MEMUSAGE=12458344
```

**Expected behavior**
kadalu-provisioner should unmount gluster volumes when not needed

**Actual behavior**
all volumes are kept mounted until restart of kadalu-csi-provisioner-0 pod

**Environment:**
<!--
Output from one of commands
1. If kubectl-kadalu plugin is installed
$ kubectl kadalu version

$ kubectl kadalu version
kubectl-kadalu plugin: 1.1.0
kadalu pod(s) versions
pod/kadalu-csi-nodeplugin-bngv6: devel
pod/kadalu-csi-nodeplugin-ftn8g: devel
pod/kadalu-csi-nodeplugin-j826m: devel
pod/kadalu-csi-nodeplugin-nb6ff: devel
pod/kadalu-csi-nodeplugin-sk2vw: devel
pod/kadalu-csi-nodeplugin-sl448: devel
pod/kadalu-csi-nodeplugin-zk289: devel
pod/kadalu-csi-provisioner-0: devel
pod/operator-5ccf95f644-h2sj7: devel

**Screenshots or Logs**
N/A

**Additional context**
N/A

madoe:
Can you check `single_pv_per_pool` in your KadaluStorage definition? 

Setting `single_pv_per_pool: true` will unmount the PV after checking its existence. (https://github.com/kadalu/kadalu/blob/d6bfe1ada056a4aa421c0c3be0f5e1ab00cbb453/csi/controllerserver.py#L224)

ramalhais-sky:
[...]

PR #1095, opened 2025-04-03
Improvement: Unmount unneeded gluster volumes on nodeserver

madoe:
Nodeserver keeps all gluster volumes mounted even if they are no longer needed, i.e. the last pvc on the node has been removed. Using many gluster volumes in the kubernetes cluster will eventually result in excessive memory usage by nodeserver. This PR adds logic to remove the PV mount in nodeserver when all PVC mounts are released. Additionally, an init system (tini) is implemented for nodeserver pods, as unmounting an gluster volume left the glusterfs process as a zombie. Tini will take care of the zombie processes.

Also includes (and supersedes) PR #1094, fixing #1083:
1. gluster volumes were not unmounted, when gluster volume name and volume name differed
2. search_volume was looping through all available volumes, instead of just the one passed as parameter, mounting all available volumes in the process.

I only could test this PR with external gluster volumes. But I believe there should be no implications with kadalu-managed volumes, as this PR only affect nodeserver.

**Checklist**
- [ ] Documentation added
- [ ] Tests updated

### DIFF
--- a/lib/kadalulib.py
+++ b/lib/kadalulib.py
@@ -132,6 +132,57 @@ def reachable_host(hosts):
             return host
     return None
 
+def get_gvolname_from_volumeid(volumeid):
+    """Returns the gluster volumename for a PVC volume ID"""
+    cmd = (
+        r'grep %s /proc/mounts '
+        r'| head -n 1 '
+        r'| cut -f 1 -d " " '
+        r'| cut -f 2 -d ":"' % (volumeid)
+    )
+
+    with subprocess.Popen(cmd,
+                          shell=True,
+                          stderr=subprocess.PIPE,
+                          stdout=subprocess.PIPE,
+                          universal_newlines=True) as proc:
+        out, err = proc.communicate()
+
+        return out.strip()
+
+def get_mntdir_from_gvolname(gvolname):
+    """Returns the mount directory for gluster volume name
+    Expectation is only one mount exists for the gluster volume"""
+    cmd = (
+        r'grep %s /proc/mounts '
+        r'| cut -f 2 -d " "' % (gvolname)
+    )
+
+    with subprocess.Popen(cmd,
+                          shell=True,
+                          stderr=subprocess.PIPE,
+                          stdout=subprocess.PIPE,
+                          universal_newlines=True) as proc:
+        out, err = proc.communicate()
+
+        return out.strip()
+
+def get_mounts_per_gvolname(gvolname):
+    """Returns how often a gluster volume is mounted"""
+    cmd = (
+        r'grep "fuse.glusterfs" /proc/mounts '
+        r'| grep ":%s " '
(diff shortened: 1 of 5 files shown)


amarts (approved):
LGTM

leelavg:
@madoe possible to sign your commits and I'll merge it?

madoe:
@leelavg Aren't they? I see a "verified" on every single commit...

leelavg:
I see DCO check failing.

madoe:
OK, commits have been PGP signed, but `Signed-off-by:` is missing in the Commit message :-(
On to it.

madoe:
@leelavg Finally signed and signed-off...

lib/kadalulib.py
```diff
@@ -66,8 +66,8 @@ def is_gluster_mount_proc_running(volname, mountpoint):
     """
     cmd = (
         r'ps ax | grep -w "/opt/sbin/glusterfs" '
-        r'| grep -w "\-\-volfile\-id %s" '
-        r'| grep -w -q "%s"' % (volname, mountpoint)
+        r'| grep -- "--volfile-id %s " '
```
leelavg: did you rebase properly, didn't we take this change already in #1092 🤔

madoe: While working on this PR I merged develop into my branch, to get the latest changes. On second thoughts a rebase would have been better :-(

Initially the merge was visible in the commits, but rebasing to add the signatures seems to have rewritten that.

madoe:
@leelavg Sorry for the delay. I recreated the PR from scratch. Hopefully everything is now OK.

lib/kadalulib.py
```diff
@@ -132,6 +132,57 @@ def reachable_host(hosts):
...
+    )
+
+    with subprocess.Popen(cmd,
+                          shell=True,
+                          stderr=subprocess.PIPE,
+                          stdout=subprocess.PIPE,
+                          universal_newlines=True) as proc:
+        out, err = proc.communicate()
+
+        return int(out)
```
leelavg: possible to construct the command at call site depending on how many times/places you are using and just use the `execute` function which already exists in this file?

basically you don't need to import every function, all you need is a `execute` func, can refer to how this is being used in https://github.com/kadalu/kadalu/blob/devel/csi/controllerserver.py#L66

madoe: This was my first approach. But using pipes in a command requires `shell=True` for `subprocess.Popen`, which is not the case in the `execute`-function.

madoe: @leelavg I modified the execute-function, so the shell flag can be passed as an argument. With this I could now remove the 3 functions I introduced earlier.

leelavg (approved):
shell=True is a bit of security issue, approving taking into account that the interpolated names are already sanitised versions either by gluster or k8s.

